PR #5, opened 2021-12-28
Add example

lu-zero:


examples/query.rs
```diff
@@ -0,0 +1,25 @@
+use mdns_sd::{ServiceDaemon, ServiceEvent};
```
keepsimple1, line 1: Could you please add a brief module doc comment at the beginning of this file? Something like:  (just a sample, you can change the wording/content).
```
//! A mDNS query client.
//! 
//! Run with:
//!
//! cargo run --example query <service_type>
//! 
```

keepsimple1 (commented):
Thanks for adding an example! I have a minor comment inline, otherwise looks good.

keepsimple1 (approved):
Thanks!

PR #9, opened 2022-01-18
Use flume instead of crossbeam-channel

lu-zero:
Makes even easier to use the library from an async context.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -18,7 +18,7 @@ jobs:
         uses: actions-rs/toolchain@v1
         with:
           profile: minimal
-          toolchain: 1.46.0
+          toolchain: 1.49.0
           override: true
           components: rustfmt, clippy
       - name: Run rustfmt and fail if any warnings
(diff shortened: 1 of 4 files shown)


.github/workflows/build.yml
```diff
@@ -18,7 +18,7 @@ jobs:
         uses: actions-rs/toolchain@v1
         with:
           profile: minimal
-          toolchain: 1.46.0
+          toolchain: 1.49.0
```
keepsimple1, line 21: I guess `flume` requires `1.49` and up?  If so, could you please also update the version number in README.md ?

Cargo.toml
```diff
@@ -10,7 +10,14 @@ keywords = ["mdns", "discovery", "service-discovery", "zeroconf", "dns-sd"]
...
+[features]
+async = ["flume/async"]
+default = ["async"]
+
 [dependencies]
-crossbeam-channel = "0.5.1" # channel between threads
 log = "0.4.14"              # logging
 nix = "0.23.0"              # socket APIs
+
+[dependencies.flume]
```
keepsimple1, line 21: Just to have a consistent style, could you please define this in the `[dependencies]` section above?

keepsimple1 (commented):
Thanks for the PR, this looks awesome! Some of the API return types would become `flume::Receiver` but the change will be worth it. I only have a couple of minor comments inline.

(A related thought is that, we don't have tests or example to cover the async use case yet. That's okay for us now because otherwise we will have to drag in dev-dependencies for async runtime, making this crate much larger. The overhead feels too big for a relatively small benefit.)

Issue #13, opened 2022-02-04
Registered services aren't showing up when using standard tools to test


dljsjr:
I'm trying to implement a very minimal service advertisement for use in some unit tests for a strictly client implementation in a different project. I'm more or less pattern matching directly off the code in the documentation; this is the current test `main` I'm using to eval the library:

```rust
fn main() {
    let dummy_service_daemon =
        mdns_sd::ServiceDaemon::new().expect("Could not create service daemon");

    let test_service_type = "_mytest._udp._local";
    let instance_name = "test_instance";
    let service_host_ipv4 = "192.168.1.176";
    let service_hostname = "192.168.1.176.local.";
    let port = 5200;

    let service_info = mdns_sd::ServiceInfo::new(
        test_service_type,
        instance_name,
        service_hostname,
        service_host_ipv4,
        port,
        None,
    );

    println!("Addresses: {:?}", &service_info.get_addresses());

    dummy_service_daemon
        .register(service_info)
        .expect("Failed to register test service");

    loop {
        std::thread::sleep(std::time::Duration::from_millis(100))
    }
}
```

The service does not seem to be advertised anywhere though. I've tested on both macOS using `dns-sd` and Ubuntu 20.04 using `mdns-scan` and the service does not show up. Some example commands I've tried:

`$> dns-sd -B _services._udp. local.` (this should show *all* dns-sd services)
`$> dns-sd -B _mytest_.udp local.` (a direct query against the service type, still shows nothing)

I'm not sure if this is a bug or if there is something missing from the docs that I haven't done correctly here. Thanks for any help you can provide with this!

keepsimple1:
>  `let test_service_type = "_mytest._udp._local";`

The type should end with ".local." .  Could you please try that?   Then you can try 

`dns-sd -B _mytest_.udp.`   or `dns-sd -B _mytest_.udp. local.` 

I've not used `-B _services._udp. local.`, will check that out.

dljsjr:
🤦 it was the `_` before `local`, which I knew shouldn't be there and I just made a copy paste bug. That said, the server still only responds to direct queries and not the "browse all" query.

dljsjr:
Also it's `_services._dns-sd._udp.local.`, I haven't had enough coffee yet (still doesn't work though)

dljsjr:
That's how I'm testing and it's not working 🤔

dljsjr:
The direct name query is fine for my unit tests, thanks!

[...]

PR #14, opened 2022-02-04
support special meta-query of dns-sd

keepsimple1:
This is to issue #13 when a wildcard browsing `_services._dns-sd._udp.local.` did not work. 

This is a missing feature. The spec is here: https://datatracker.ietf.org/doc/html/rfc6763#section-9. 

Test and example code are added.


keepsimple1:
It is strange that the new test only fails in GitHub and succeeds in local macOS and Linux.

Issue #18, opened 2022-02-27
add Windows support


keepsimple1:
NOTE: if you are familiar with socket support on Windows in Rust, please feel free to comment on this issue. 

I started to find cases when I wanted to use this library on Windows. Don't know how to implement yet. This is a placeholder to this new feature.

Update: after some research I plan to use `winapi` crate to add support on Windows. Hopefully there are not too many gotcha.

keepsimple1:
@indietyp @lu-zero  I wonder if any of you are interested in Windows support of this lib?  If yes, what do you think of the approach I tried in the linked PR?

lu-zero:
I do not use windows, so I cannot say much sadly :/

indietyp:
I have a windows machine (which I do not use primarily for development), so I'd be happy to take a look.

keepsimple1:
The PR is merged now.

PR #19, opened 2022-03-19
add support on Windows

keepsimple1:
see issue #18 . 

- Use `socket2` for cross-platform socket programming. 
- Use `polling` to support cross-platform polling / select.

If successful, will no longer use `nix`.

~~Pending considerations:~~
- `socket2` now has two branches `0.3` and `0.4`, and they use some different approaches.  We are using `0.3` branch because its socket [`recv_from`](https://docs.rs/socket2/0.3.19/socket2/struct.Socket.html#method.recv_from) method has better ergonomics, while `0.4` branch [ `recv_from`](https://docs.rs/socket2/0.4.4/socket2/struct.Socket.html#method.recv_from) uses `MayebUninit` which is hard to use. 

Update 2022-6-13:

- All things considered, I moved to the latest `socket2` (i.e. branch 0.4) because it will be easier to update / track in the future.  The trade-off is to use `Read` trait instead of `recv_from`, andwe will no longer log the peer address.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,5 +17,5 @@ default = ["async"]
 [dependencies]
 flume = { version = "0.10", default-features = false }     # channel between threads
 log = "0.4.14"                                             # logging
-socket2 = { version = "0.4", features = ["all"] }          # socket APIs
 polling = "2.1.0"                                          # select/poll sockets
+socket2 = { version = "0.4", features = ["all"] }          # socket APIs
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,6 +15,7 @@ async = ["flume/async"]
 default = ["async"]
 
 [dependencies]
-flume = { version = "0.10", default-features = false } # channel between threads
-log = "0.4.14"                                         # logging
-nix = "0.24.1"                                         # socket APIs
+flume = { version = "0.10", default-features = false }     # channel between threads
+log = "0.4.14"                                             # logging
+socket2 = { version = "0.3.19", features = ["reuseport"] } # socket APIs
+polling = "2.1.0"                                          # select/poll sockets
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,5 +17,5 @@ default = ["async"]
 [dependencies]
 flume = { version = "0.10", default-features = false }     # channel between threads
 log = "0.4.14"                                             # logging
-socket2 = { version = "0.4.4", features = ["all"] }        # socket APIs
+socket2 = { version = "0.4", features = ["all"] }          # socket APIs
 polling = "2.1.0"                                          # select/poll sockets
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,5 +17,5 @@ default = ["async"]
 [dependencies]
 flume = { version = "0.10", default-features = false }     # channel between threads
 log = "0.4.14"                                             # logging
-socket2 = "0.4.4"                                          # socket APIs
+socket2 = { version = "0.4.4", features = ["all"] }        # socket APIs
 polling = "2.1.0"                                          # select/poll sockets
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,5 +17,5 @@ default = ["async"]
 [dependencies]
 flume = { version = "0.10", default-features = false }     # channel between threads
 log = "0.4.14"                                             # logging
-socket2 = { version = "0.3.19", features = ["reuseport"] } # socket APIs
+socket2 = "0.4.4"                                          # socket APIs
 polling = "2.1.0"                                          # select/poll sockets
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -143,7 +143,8 @@ use std::{
     convert::TryInto,
     fmt,
     net::{Ipv4Addr, SocketAddrV4},
-    str, thread,
+    str::{self, FromStr},
+    thread,
     time::{Duration, SystemTime},
     vec,
 };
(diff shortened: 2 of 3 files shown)


anenankhu:
Tried running examples against this branch on Windows 11,  browse example fires SearchStarted and SearchFound events, but never a ServiceResolved event

keepsimple1:
> Tried running examples against this branch on Windows 11, query example fires SearchStarted and SearchFound events, but never a ServiceResolved event
> 

I tried again on Windows 11, and it still works for me.  Could you please copy and paste the output of `cargo run --example query <your-service>` ? 

Note that in the example code ServiceResolved event is not printed with the event name.  The log shows "Resolved a new service ..."

Cargo.toml
```diff
@@ -15,6 +15,7 @@ async = ["flume/async"]
 default = ["async"]
 
 [dependencies]
-flume = { version = "0.10", default-features = false } # channel between threads
-log = "0.4.14"                                         # logging
-nix = "0.23.1"                                         # socket APIs
+flume = { version = "0.10", default-features = false }     # channel between threads
+log = "0.4.14"                                             # logging
+socket2 = { version = "0.3.19", features = ["reuseport"] } # socket APIs
```
keepsimple1, line 20: Unfortunately, `socket2` seems to split into two branches now:  `0.3` and `0.4`.  We opt to use `0.3` for now mainly due to its socket [`recv_from`](https://docs.rs/socket2/0.3.19/socket2/struct.Socket.html#method.recv_from) method .

keepsimple1:
@anenankhu I rebased this branch to the latest main branch release 0.5.0, and tested again on Windows 11.  Both the client side query and the service side publish seems to be working.  If you see any issues, please let me know. Thanks. 

The main reason that I have not merged in this PR is the pending consideration about `socket` lib that I mentioned in Description.

Cargo.toml
indietyp: There would be a `Read` trait implementation (for `read`) which we could adapt: https://github.com/rust-lang/socket2/blob/master/src/socket.rs#L1816 if we wanted to go with `socket2` `0.4` support.

keepsimple1: `read` of `Read` trait calls `recv`, hence returns only the buffer, not the source address.  Currently we are using `recv_from` which returns both the buffer and the source address.  It's true that we are not utilizing the source address at the moment (besides logging) as we only do multicast, but I'm not sure if we want it for future or not. It will be a trade-off.

indietyp: I am sorry for the delay, kind of... forgot, what I mean is that we could use the function that already does it via the trait and adapt it.

```rust
fn read_from(sock: &mut Socket), buf: &mut [u8]) -> IpAddr {
        let buf = unsafe { &mut *(buf as *mut [u8] as *mut [MaybeUninit<u8>]) };
        self.recv_from(buf)
}
```

This is more pseudo-code than anything.

keepsimple1: no problem. I didn't want to use `unsafe` in this code base though. In comparison,  changing `recv_from` to `read` seems to strike a better trade-off.

keepsimple1: I finally moved to the current `socket2` release (i.e. 0.4 branch) to be future-proof. I felt it's the better trade-off.  And I plan to merge these changes if no other issues.

anenankhu:
@keepsimple1 
tried latest, doesn't work: 
register: 

```
> cargo.exe run -q --example register _abc._tcp.local. test
Registered service test._abc._tcp.local.
```

query:
```
>
cargo.exe run -q --example query _abc._tcp
Received other event: SearchStarted("_abc._tcp.local.")
Received other event: ServiceFound("_abc._tcp.local.", "test._abc._tcp.local.")
Received other event: SearchStarted("_abc._tcp.local.")
Received other event: ServiceFound("_abc._tcp.local.", "test._abc._tcp.local.")
Received other event: SearchStarted("_abc._tcp.local.")
Received other event: ServiceFound("_abc._tcp.local.", "test._abc._tcp.local.")
Received other event: SearchStarted("_abc._tcp.local.")
Received other event: ServiceFound("_abc._tcp.local.", "test._abc._tcp.local.")
```
As you can see, there isn't ServiceResolved event in the output, only ServiceFound.

anenankhu:
If that helps, here's my windows verstion: 
image

Also rust version:  
```
> rustc.exe -vV
rustc 1.63.0-nightly (ee160f2f5 2022-05-23)
binary: rustc
commit-hash: ee160f2f5e73b6f5954bc33f059c316d9e8582c4
commit-date: 2022-05-23
host: x86_64-pc-windows-msvc
release: 1.63.0-nightly
LLVM version: 14.0.4
```

keepsimple1:
@anenankhu  thank you very much for trying it out.  I only tested using my use cases and didn't test `examples` in our repo. Your test revealed a bug I missed, i.e. we only resolve the service instances if the `properties` is not empty. That is too restrictive and our example `register.rs` [did not set any properties](https://github.com/keepsimple1/mdns-sd/blob/8e863f456d93af8893c5dd005481cd0b6bc25c8d/examples/register.rs#L34). Hence the failure.  I just updated the diff to fix this and now it should work for you.  Please let me know if it works or not.  Thanks!

anenankhu:
Actually, as long as it resolves you can always start over when it's necessary to get fresh address, so that's fine. It seems to be working real well on windows, so when will there be a merge? 😄 I'm more than excited to finally use `mdns-sd` in my project!

keepsimple1:
Thanks @anenankhu for confirming. Yes, the current design is that we send 'resolve' of an instance only once during one 'browse'. The caller would 'browse' again to get a new resolve for the same instance. If we find problems in future, we can always revisit this logic.

I will merge this PR for now.

Issue #20, opened 2022-05-09
allow multiple addresses in ServiceInfo


indietyp:
The `ServiceInfo` struct can hold multiple corresponding addresses, but there is no way to add more than a single addr (only one via `ServiceInfo::new`). Would it be possible to add the ability to define multiple addresses?

keepsimple1:
Since `host_ipv4` is `&str` in `ServiceInfo::new`,  we could add support specifying one or more ipv4 addresses in the same string, separated by `,` or some other separator.  What do you think?  (any other suggestions are welcome).

indietyp:
Going with a string separated by `,` would be a good idea, but also adding methods like `insert_address` and `remove_address` for a more programmatic approach.

One could also imagine having an optional dependency on the [ipnet](https://crates.io/crates/ipnet) crate, which provides access to an [`Ipv4AddrRange`](https://docs.rs/ipnet/2.5.0/ipnet/struct.Ipv4AddrRange.html) and [`Ipv4Net`](https://docs.rs/ipnet/2.5.0/ipnet/struct.Ipv4Net.html) struct.

I'd be happy to create a pull request that implements those changes.

keepsimple1:
Adding methods like `insert_address` and `remove_address` sounds good to me.  I'm reluctant to add a new dependency for this purpose, unless something really useful.

Thank you for submitting the PR. Will take a look.

indietyp:
#21 was merged, closing this issue

PR #21, opened 2022-05-10
allow multiple formats for host_ipv4

indietyp:
This pull request implements the issue raised in #20 and creates a new trait: `AsAddr`, and a new optional dependency, `ipnet`, to offer greater interoperability.

This change is backwards compatible. `host_ipv4` now supports the following additional formats:

* `&str` of ip (behaviour from before the pull)
* `String` of ip
* comma-separated `&str` or `String` of ip (e.g. `127.0.0.1,127.0.0.2`)
* `std::net::Ipv4Addr`
* `nix::sys::socket::Ipv4Addr`
* (feature: ipnet) comma-separated `&str` or `String` of ip or CIDR (e.g. `127.0.0.0/24,127.2.0.0`)
* (feature: ipnet) `ipnet::Ipv4Net`
* (feature: ipnet) `ipnet::Ipv4AddrRange`
* `&T` of any of the above
* `&[T]` of any of the above

We should consider if it might be worth it to convert `ServiceInfo::new()` to return a result so that parse errors aren't silently removed.

This pull request also adds a new Error variant called `Error::ParseIpAddr`


src/lib.rs
```diff
@@ -1365,6 +1364,109 @@ impl DnsCache {
...
+                        .and_then(|net| net.as_addr()),
+                    #[cfg(not(feature = "ipnet"))]
+                    Err(err) => Err(Error::ParseIpAddr(err.to_string())),
+                }
+            })
+            // this collect isn't ideal, but helps us remove the result err
+            .collect::<Result<Vec<_>>>()?
+            .into_iter()
+            .flatten()
+            .collect();
```
keepsimple1, line 1403: I don't think you really need these `collect` calls. As mentioned earlier, you can new a `HashSet` first, then fill it up in a simple loop or a `map`.

src/lib.rs
```diff
@@ -1365,6 +1364,109 @@ impl DnsCache {
...
+
+impl<I: AsAddr> AsAddr for &[I] {
+    fn as_addr(&self) -> Result<HashSet<Ipv4Addr>> {
+        let ips = self
+            .into_iter()
+            .map(|val| val.as_addr())
+            .collect::<Result<Vec<HashSet<_>>>>()?
+            .into_iter()
+            .flatten()
+            .collect();
```
keepsimple1, line 1423: I would suggest creating a new empty `HashSet` first, then use a simple loop to iterator the `&[T]` and fill in the new set.

src/lib.rs
```diff
@@ -1365,6 +1364,109 @@ impl DnsCache {
     }
 }
 
+pub trait AsAddr {
+    fn as_addr(&self) -> Result<HashSet<Ipv4Addr>>;
```
keepsimple1, line 1368: To make it clear, can we name this method `as_ipv4_addr_set`? or anything to convey that meaning? 

So that, if this trait supports more address family or data structures in future, it will be easy to add. 

If we don't plan to support any new address or data structure in future for this trait, then I would suggest to name the trait itself to `AsIPv4AddrSet` or something like that.

src/lib.rs
```diff
@@ -1454,6 +1556,16 @@ impl ServiceInfo {
         &self.addresses
     }
 
+    /// Insert an additional address to the service
+    pub fn insert_address(&mut self, addr: Ipv4Addr) {
+        self.addresses.insert(addr);
+    }
+
+    /// Remove an address
+    pub fn remove_address(&mut self, addr: &Ipv4Addr) {
```
keepsimple1, line 1565: same here, can we name it `remove_ipv4`?

keepsimple1 (commented):
Thanks for your PR! Please see comments inline.

indietyp:
I thought ipnet might be a good convenience feature (for my application it would simplify some things)

I can remove the ipnet dependency and code.

src/lib.rs
```diff
@@ -132,45 +132,44 @@
 // In mDNS and DNS, the basic data structure is "Resource Record" (RR), where
 // in Service Discovery, the basic data structure is "Service Info". One Service Info
 // corresponds to a set of DNS Resource Records.
+use std::any::Any;
```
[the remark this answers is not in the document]
indietyp: Yes of course, this must have slipped, my rustfmt configuration automatically sorts and normalizes imports

src/lib.rs
```diff
@@ -1454,6 +1556,16 @@ impl ServiceInfo {
         &self.addresses
     }
 
+    /// Insert an additional address to the service
+    pub fn insert_address(&mut self, addr: Ipv4Addr) {
```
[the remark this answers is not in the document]
indietyp: sure, will do

src/lib.rs
indietyp: sure, will do.

src/lib.rs
indietyp: maybe `as_ipv4_addrs`?

src/lib.rs
```diff
@@ -132,45 +132,44 @@
...
+#[derive(Debug, PartialEq)]
 pub enum Error {
     /// Like a classic EAGAIN. The receiver should retry.
     Again,
 
     /// A generic error message.
     Msg(String),
+
+    /// Error during parsing of ip address
+    ParseIpAddr(String),
```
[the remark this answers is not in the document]
indietyp: are you sure? parsing the ip addr from string can result in an error, right now that error is just ignored, which I think could be surprising for new users.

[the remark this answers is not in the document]
keepsimple1: I think this will be useful if we change `ServiceInfo::new()` to return a `Result` (You mentioned in another place as well).

I'm wondering if we should take this opportunity to change `new()`, which is a breaking change of course.

src/lib.rs
keepsimple1: Sure, `as_ipv4_addrs` sounds good.  And I think it looks more consistent and concise if the trait name is also changed to `AsIpv4Addrs` .  

A trait with a single method feels right in this case (which is what you have here).  If we ever need to support IPv6, it's probably better to define a new trait instead of adding new methods into this trait.

indietyp:
(just a small note: I will be able to implement all those changes this week, hopefully sooner or later, sorry for the potential delay)

src/lib.rs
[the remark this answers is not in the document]
indietyp: I'd be all for changing `new()`, even if it is a breaking change. I imagine new users might feel very frustrated and/or irritated if we just silence IP parsing errors.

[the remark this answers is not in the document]
indietyp: prob. offtopic here, but have you also thought about some reorganization of the project. Getting into the project was quite hard, as there's just a single 2.5k `lib.rs`. I think it would be beneficial for maintainability (and new features?) if we would split up the project into separate files. What do you think?

src/lib.rs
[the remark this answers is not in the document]
keepsimple1: Btw, since you defined `AsIpv4Addrs`,  can you use that trait for this method?  

As `Ipv4Addr` in this lib currently is from `Nix` crate, and I wanted to avoid exposing data structure from `Nix` directly to our caller.  Using a trait instead would be useful for the caller, similar like in `new()`.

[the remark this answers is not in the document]
indietyp: I thought it might have been a good addition when you want to manipulate the `ServiceInfo` after `::new()`, but before registering it, but now that I think about it, I don't think there's a use-case for this functionality without a new command for the daemon.

indietyp:
@keepsimple1 I updated the PR, hopefully it addressed all of the problems

Cargo.toml
```diff
@@ -17,4 +17,4 @@ default = ["async"]
 [dependencies]
 flume = { version = "0.10", default-features = false } # channel between threads
 log = "0.4.14"                                         # logging
-nix = "0.23.1"                                         # socket APIs
+nix = "0.23.1"                                         # socket APIs
```
keepsimple1, line 20: nitpick:  not a real change, need to remove it.

src/lib.rs
```diff
@@ -912,6 +917,14 @@ impl Zeroconf {
             for record in records.iter() {
                 if let Some(ptr) = record.any().downcast_ref::<DnsPointer>() {
                     let info = self.create_service_info_from_cache(ty_domain, &ptr.alias);
+                    let info = match info {
+                        Ok(ok) => ok,
+                        Err(err) => {
+                            error!("Error while creating service info from cache: {:?}", err);
```
keepsimple1, line 923: nitpick:  `{}` works for `err`.

src/lib.rs
```diff
@@ -1033,10 +1050,18 @@ impl Zeroconf {
...
+                let service_info = ServiceInfo::new(&service_type, &my_name, "", (), 0, None);
+
+                match service_info {
+                    Ok(service_info) => {
+                        debug!("Inserting service info: {:?}", &service_info);
+                        self.instances_to_resolve
+                            .insert(instance.clone(), service_info);
+                    }
+                    Err(err) => {
+                        error!("Malformed service info while inserting: {:?}", err);
```
keepsimple1, line 1062: nitpick:  `{}` works for `err`.

tests/addr_parse.rs
```diff
@@ -0,0 +1,68 @@
+#[cfg(feature = "ipnet")]
+use ipnet::{Ipv4AddrRange, Ipv4Net};
```
keepsimple1, line 2: `ipnet` is no longer needed.

src/lib.rs
```diff
@@ -152,25 +152,30 @@ use std::{
     convert::TryInto,
     fmt,
     os::unix::io::RawFd,
+    str::FromStr,
```
keepsimple1, line 155: I think this format would cause  `cargo fmt` failure.

src/lib.rs
```diff
@@ -152,25 +152,30 @@ use std::{
...
     str, thread,
     time::SystemTime,
     vec,
 };
 
 /// A basic error type from this library.
-#[derive(Debug)]
+#[derive(Debug, PartialEq)]
```
keepsimple1, line 162: I was thinking if we should use `#[non_exhaustive]` here.  I don't have enough experience to make a definitive choice. I will leave as is.

keepsimple1 (commented):
Thanks for updating! Overall looks good to me. Only a few minor comments.

src/lib.rs
indietyp: good catch, error variants definitely should be `#[non_exhaustive]`

src/lib.rs
```diff
@@ -1365,6 +1391,79 @@ impl DnsCache {
...
+
+/// Optimization for zero sized/empty values, as `()` will never take up any space or evaluate to
+/// anything, helpful in contexts where we just want an empty value.
+impl AsIpv4Addrs for () {
+    fn as_ipv4_addrs(&self) -> Result<HashSet<Ipv4Addr>> {
+        Ok(HashSet::new())
+    }
+}
+
+impl<I: AsIpv4Addrs, const N: usize> AsIpv4Addrs for [I; N] {
```
keepsimple1, line 1443: const generics is stable after Rust 1.51, but currently we support earliest to Rust 1.49.  How much benefit this support of fixed array bring? 

Since you already support slice `&[I]`,  it seems to me that we don't really need to implement this trait for fixed array directly.  

But if I missed something and there is indeed use case for this,  I will be OK to bump up required Rust version to 1.51.  

Btw, I didn't find how to allow your PR to run GitHub action workflow automatically.  It's probably only a first PR thing, I will run it as soon as I see it.  Sorry about that delay.

tests/addr_parse.rs
```diff
@@ -0,0 +1,66 @@
...
+            Ipv4Addr::from_std(&std::net::Ipv4Addr::new(127, 0, 0, 2))
+        ]
+        .into())
+    );
+}
+
+#[test]
+fn test_addr_slice() {
+    assert_eq!(
+        ["127.0.0.1"].as_slice().as_ipv4_addrs(),
```
keepsimple1, line 30: array_as_slice is stable in Rust 1.57.  But it is equivalent to `&s[..]`.  Is it OK for us just use `&s[..]` ?

keepsimple1 (commented):
We're almost ready to merge. Only a couple of comments left due to workflow failures and Rust version.

tests/addr_parse.rs
indietyp: Yea, we can use it. I completely forgot that we have `1.49` as our minimum version.

src/lib.rs
indietyp: Thanks for all the feedback. This was more or less a by-product of my experimentation to easily allow an empty value (which was solved with the `()` implementation). You are entirely correct, with `&[T]` already being implemented, `[T; N]` is redundant

indietyp:
my mistake with the tests, fixing them now

tests/addr_parse.rs
```diff
@@ -0,0 +1,97 @@
...
+            set.insert(Ipv4Addr::from_std(&std::net::Ipv4Addr::new(127, 0, 0, 1)));
+
+            set
+        })
+    );
+
+    let addr = "127.0.0.1".to_string();
+    assert_eq!(
+        addr.as_ipv4_addrs(),
+        Ok(HashSet::from_iter([Ipv4Addr::from_std(
```
keepsimple1, line 22: Until Rust 1.53, only references to arrays implement IntoIterator ([doc](https://doc.rust-lang.org/edition-guide/rust-2021/IntoIterator-for-arrays.html)). You should be able to use `(&[array]).into_iter()` .

indietyp: o yea, I fixed it in the other examples via an explicit construction of the `HashSet`, but I must've overlooked that one

keepsimple1:
Good work, thanks!

PR #23, opened 2022-05-17
impl AsIpv4Addrs refactoring

keepsimple1:
When trying to use the latest code in `main`, I found that it missed `impl AsIpv4Addrs` trait for `&String`, which I used with previous versions. 

To support `AsIpv4Addrs` with reference types and avoid redundancy,  this diff changes the rest of `impl`s to use reference types.

This is a follow-up to PR #21 .  @indietyp  any comments/thoughts?

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1395,6 +1395,12 @@ pub trait AsIpv4Addrs {
     fn as_ipv4_addrs(&self) -> Result<HashSet<Ipv4Addr>>;
 }
 
+impl<T: AsIpv4Addrs> AsIpv4Addrs for &T {
+    fn as_ipv4_addrs(&self) -> Result<HashSet<Ipv4Addr>> {
+        (*self).as_ipv4_addrs()
+    }
+}
+
 impl AsIpv4Addrs for &str {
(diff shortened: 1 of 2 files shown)


indietyp:
I must've deleted the implementation, but there was an original implementation. (https://github.com/keepsimple1/mdns-sd/pull/21/commits/f2627ac79a2f7b880702f2071cf08687bad360d2#diff-b1a35a68f14e696205874893c07fd24fdb88882b47c23cc0e0c80a30c7d53759R1370)

Tho, it would have been:

```rust
impl<T: AsIpv4Addrs> for &T {
    fn as_ipv4_addrs(&self) {
       T::as_ipv4_addrs(self)
    }
}
```

or something similar.

indietyp:
I just checked, I must've removed the implementation by accident in [`70443c12309a5c0e014696f9a0795dfcffea038d`](https://github.com/keepsimple1/mdns-sd/pull/21/commits/70443c12309a5c0e014696f9a0795dfcffea038d).

keepsimple1:
Thanks for bringing that up. That's a better approach!  I updated the diff.

PR #44, opened 2022-07-09
Join all interfaces to multicast

fufesou:
Fix multicast dgram listen on a multiple net interfaces platform.

If call [`join_multicast_v4()`](https://docs.rs/socket2/0.4.4/socket2/struct.Socket.html#method.join_multicast_v4:~:text=If%20it%E2%80%99s%20Ipv4Addr%3A%3AUNSPECIFIED%20(INADDR_ANY)%20then%20an%20appropriate%20interface%20is%20chosen%20by%20the%20system.) with `0.0.0.0` on Win10, a random interface will be choose. The `listen_socket` may not be able to receive multicast dgram.

Signed-off-by: fufesou <shuanglongchen@yeah.net>

### DIFF
--- a/src/service_daemon.rs
+++ b/src/service_daemon.rs
@@ -520,13 +520,14 @@ impl Zeroconf {
         debug!("created listening socket: {:?}", &listen_socket);
 
         let group_addr = Ipv4Addr::new(224, 0, 0, 251);
-        listen_socket
-            .join_multicast_v4(&group_addr, &Ipv4Addr::new(0, 0, 0, 0))
-            .map_err(|e| e_fmt!("join multicast group: {}", e))?;
 
         // We create a socket for every outgoing IPv4 interface.
         let mut respond_sockets = Vec::new();
         for ipv4_addr in my_ipv4_addrs() {
+            listen_socket
+                .join_multicast_v4(&group_addr, &ipv4_addr)
+                .map_err(|e| e_fmt!("join multicast group: {}", e))?;
+
             let respond_socket = new_socket(ipv4_addr, udp_port, false)?;
             respond_sockets.push(respond_socket);
         }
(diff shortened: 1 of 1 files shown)


keepsimple1:
Thanks for opening the PR! Is this an issue only on Win10?  How about Win11?

fufesou:
> Is this an issue only on Win10? How about Win11?

I'll try to install a Win11 VM and test it.
By the way, [webrtc-rs/mdns](https://github.com/webrtc-rs/mdns/blob/0fbe2aa92beb8d75014e18c6b9ea5aa44aacb710/src/conn/mod.rs#L79) also joins all interfaces.

src/service_daemon.rs
```diff
@@ -520,13 +520,14 @@ impl Zeroconf {
...
-        listen_socket
-            .join_multicast_v4(&group_addr, &Ipv4Addr::new(0, 0, 0, 0))
-            .map_err(|e| e_fmt!("join multicast group: {}", e))?;
 
         // We create a socket for every outgoing IPv4 interface.
         let mut respond_sockets = Vec::new();
         for ipv4_addr in my_ipv4_addrs() {
+            listen_socket
+                .join_multicast_v4(&group_addr, &ipv4_addr)
+                .map_err(|e| e_fmt!("join multicast group: {}", e))?;
```
keepsimple1, line 529: nit pick:  could you log the `&ipv4_addr` in the message as well?  For example:
```
e_fmt!("join multicast group on addr {}: {}", &ipv4_addr, e)
```

keepsimple1 (commented):
Looks good.  I didn't know we can join multicast group on multiple interfaces using the same socket. One minor comment inline.

keepsimple1:
> I'll try to install a Win11 VM and test it. 

Now I understand more, and this looks like a good fix. Win11 is not a blocking issue.

src/service_daemon.rs
fufesou: Ok

keepsimple1 (approved):

Issue #46, opened 2022-07-11
Should the daemon send the preset IPs to the browser


fufesou:
Suppose we have some lans like the following:

![image](https://user-images.githubusercontent.com/13586388/178173663-91021753-ac43-437c-b367-ed5f84f90826.png)

`B` send a query. If `A`'s response contains `172.21.5.234`, it will be confusing.
In 'A''s response, the IP may be better set to the binding socket.

And I also think response may be better to send back to the interface, witch received the query. No need to send outging responses to all interfaces.

keepsimple1:
> And I also think response may be better to send back to the interface, witch received the query. No need to send outging responses to all interfaces.

I agree. It was a [trade-off](https://github.com/keepsimple1/mdns-sd/pull/19#discussion_r895961974) that we moved to use `socket2` current release that does not have a good safe `recv_from` API. Let me think about a bit. 

Also feel free to share your thoughts regarding how we can fix or work around this.

keepsimple1:
@fufesou To figure out what we should do for a host that resides in more than one LAN: 

1. Query: do you want to send out the query of service `foo` in all connected LANs? 

2. If the answer of 1 is yes,  what happens if there are responses from two LANs with the same service instance names, but different IP address? In such case, do you want to merge them into the one `ServiceInfo` back to the client, or have two separate `ServiceInfo` ?   (I'm still trying to find related info in RFC 6762).

3. Response: if the connect LANs can route packets between them,  do you want hosts in LAN A be aware of service instances in LAN B? 

4. If the answer of 3 is no,  do you think it's OK to check the DNS SRV record address to figure out what outgoing interface to send?  For example, if the SRV record has address 192.168.1.2,  we will only send the response out to the interface that in the LAN of 192.168.1.0/24.    (If we do this,  we can avoid dependency of `recv_from` API in `socket2`, which is unsafe in its current release). 

also cc @lu-zero @indietyp  as I am interested in their thoughts too, and if they have seen related use cases.

fufesou:
My mistake, I want to show this picture.
![image](https://user-images.githubusercontent.com/13586388/178381318-c35c9165-0a13-4eba-8a36-f280a26e24dc.png)

1. Answer of 1 is yes. I think sending two separate `ServiceInfo` is good. Merging into one `ServiceInfo` may cause confusing.
[...]

Issue #47, opened 2022-07-15
Join multicast group error: An invalid argument was supplied. (os error 10022)


fufesou:
Log from [mikekgr](https://github.com/mikekgr) in discussion https://github.com/rustdesk/rustdesk/discussions/844#discussioncomment-3111117

OS: Win10
```
IPv4 addrs: [100.125.199.128, 192.168.0.108, 192.168.1.186, 172.29.96.1]
new socket bind to 100.125.199.128:5353
new socket bind to 192.168.0.108:5353
join multicast group on addr 192.168.1.186: An invalid argument was supplied. (os error 10022), 
```

The [code](https://github.com/keepsimple1/mdns-sd/blob/23ed4acb9432efae922eeb6febb328a0c6968f89/src/service_daemon.rs#L527).

I can't reproduce on my PC. I'm not sure if the problem is `100.125.199.128`.
I'll contact him later for more tests.

keepsimple1:
> join multicast group on addr 192.168.1.186: An invalid argument was supplied. (os error 10022), 

It seems that Windows has some special remarks in their [docs](https://docs.microsoft.com/en-us/windows/win32/winsock/multicast-socket-option-behavior): 

> Do not call IP_ADD_MEMBERSHIP with the same group more than once on the same network interface.

But the error code does not match (`WSAEADDRNOTAVAIL` 10049 vs `WSAEINVAL` 10022). And it's not clear what `the same group` means in Windows.  So I'm not sure, just a thought.

keepsimple1:
After the above PR #48 merged, could you please check if this issue still exists?

fufesou:
Ok, I'll contant [mikekgr](https://github.com/mikekgr) for help. Also I'll ask him if his email can be shared to you.

keepsimple1:
I don't think this problem exist any longer. If you found otherwise, please re-open this issue, or create a new issues. Thanks!

PR #48, opened 2022-07-17
track IPv4 interfaces with sockets to support multiple LANs

keepsimple1:
Fix the issue #46 .   (It might also help fixing #47, need testing ) 

- Use a new struct `IntfSock` to link the interface with the socket.
- Only include the DNS `A` record of the same LAN as the outgoing interface.
- Use the same socket for receiving and sending on an interface.

### DIFF
--- a/examples/register.rs
+++ b/examples/register.rs
@@ -9,7 +9,9 @@
 //! cargo run --example register _my-hello._udp.local. test1
 //!
 
+use if_addrs::{IfAddr, Ifv4Addr};
 use mdns_sd::{ServiceDaemon, ServiceInfo};
+use std::net::Ipv4Addr;
 
 fn main() {
     // Create a new mDNS daemon.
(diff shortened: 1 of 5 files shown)


fufesou:
https://github.com/keepsimple1/mdns-sd/issues/46#issuecomment-1186483112
If `192.168.100.103` is not in [adresses](https://github.com/keepsimple1/mdns-sd/blob/946f04307e98bc1190cac68fba21f5968a5a87d0/src/service_info.rs#L25). A query from `192.168.100.102` will also be answered. The register responses to the browser without ip.
I'm not sure if it was intentional.

keepsimple1:
I think this is the case of responding PTR query. I was not sure if should skip the service if no addrs in the LAN of interface. Now I think it's better to skip such service, i.e. not to send out answers if no address. This is also consistent with the announcing logic. I've updated the diff to reflect this. Let me know if makes sense / works for you.

fufesou:
Great job! It works now.

keepsimple1:
Thanks for helping test.

PR #56, opened 2022-08-11
Allow service names with trailing '.'

dalepsmith:
trim_end_matches('.') removes all trailing '.', so use
strip_suffix('.') instead.


keepsimple1:
Refer to issue #57  .

keepsimple1 (approved):
Thanks for the PR. Looks good to me.

Issue #63, opened 2022-09-05
Make log dependency optional


pixsperdavid:
If the log crate is being used for logging in a downstream project, messages from mdns-sd can be undesirable. To avoid this, the dependency could be made optional. 

Here's an example on how this works in the rustls crate:
https://github.com/rustls/rustls/blob/86f4695c8c05a8041807a34d691b9bb07b757df0/rustls/src/lib.rs#L302

Happy to open a PR to implement this if acceptable.

keepsimple1:
Sounds like a good improvement. Thanks for opening the issue and the PR.

pixsperdavid:
Implemented in #64

PR #64, opened 2022-09-05
Changed dependency on 'log' crate to be optional

pixsperdavid:
As suggested in #63

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,11 +12,12 @@ description = "mDNS Service Discovery library with no async runtime dependency"
 
 [features]
 async = ["flume/async"]
-default = ["async"]
+default = ["async", "logging"]
+logging = ["log"]
 
 [dependencies]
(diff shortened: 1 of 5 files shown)


Cargo.toml
```diff
@@ -12,11 +12,12 @@ description = "mDNS Service Discovery library with no async runtime dependency"
 
 [features]
 async = ["flume/async"]
-default = ["async"]
+default = ["async", "logging"]
```
keepsimple1, line 15: Could you please move `default` line towards after `logging` since `default` is special and I think it's better to be the last line of this section.

keepsimple1 (commented):
Thanks for the PR. I have only a minor comment, and there are some minor cargo fmt errors to fix.

pixsperdavid:
Done, and fmt errors fixed.

keepsimple1 (approved):
LGTM!

PR #74, opened 2022-11-11
Ignore duplicate keys

dalepsmith:
https://github.com/keepsimple1/mdns-sd/issues/74

### DIFF
--- a/src/service_info.rs
+++ b/src/service_info.rs
@@ -302,7 +302,7 @@ fn decode_txt(txt: &[u8]) -> HashMap<String, String> {
                 Some(idx) => {
                     let k = &kv_string[..idx];
                     let v = &kv_string[idx + 1..];
-                    kv_map.insert(k.to_string(), v.to_string());
+                    kv_map.entry(k.to_string()).or_insert_with(|| v.to_string());
                 }
                 None => error!("cannot find = sign inside {}", &kv_string),
             },


dalepsmith:
Actually #74

keepsimple1:
> #74

Shouldn't it be a link to the issue?

src/service_info.rs
```diff
@@ -302,7 +302,7 @@ fn decode_txt(txt: &[u8]) -> HashMap<String, String> {
                 Some(idx) => {
                     let k = &kv_string[..idx];
                     let v = &kv_string[idx + 1..];
-                    kv_map.insert(k.to_string(), v.to_string());
+                    kv_map.entry(k.to_string()).or_insert_with(|| v.to_string());
```
keepsimple1, line 305: nit: why not use `or_insert(v.to_string())`?

keepsimple1 (commented):
Looks good. Thanks! A nit inline.

dalepsmith:
On 11/11/22, keepsimple1 ***@***.***> wrote:
> @keepsimple1 commented on this pull request.
>
> Looks good. Thanks! A nit inline.
>
>> @@ -302,7 +302,7 @@ fn decode_txt(txt: &[u8]) -> HashMap<String, String>
>> {
>                  Some(idx) => {
>                      let k = &kv_string[..idx];
>                      let v = &kv_string[idx + 1..];
> -                    kv_map.insert(k.to_string(), v.to_string());
> +                    kv_map.entry(k.to_string()).or_insert_with(||
> v.to_string());
>
> nit: why not use `or_insert(v.to_string())`?

Because Clippy didn't like it.  ;^}

WIth `or_insert`, the `to_string` is called every time, if it's needed
or not. With `or_insert_with`, `to_string` is *only* called when
actually doing the insert.

-Dale

dalepsmith:
On 11/11/22, keepsimple1 ***@***.***> wrote:
>> #74
>
> Shouldn't it be a link to the issue?

Sorry.  Still new at doing this kind of thing though the github web interface.

-Dale

keepsimple1:
> WIth `or_insert`, the `to_string` is called every time, if it's needed
or not. With `or_insert_with`, `to_string` is *only* called when
actually doing the insert.

Good point! Indeed, such pattern is quite common in Rust.

keepsimple1 (approved):

PR #77, opened 2022-11-12
Use CaseInsesitiveHashMap for TXT keys

dalepsmith:
The RFC requires TXT keys to be case insensitive.

This simply replaces `HashMap<String, String>`
with `CaseInsensitiveHashMap<String>`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,6 +16,7 @@ logging = ["log"]
 default = ["async", "logging"]
 
 [dependencies]
+case_insensitive_hashmap = "1.0.0"
 flume = { version = "0.10", default-features = false } # channel between threads
 if-addrs = "0.7"                                       # get local IP addresses
 log = { version = "0.4.14", optional = true }          # logging
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -62,7 +62,7 @@
 //!
 //! ```rust
 //! use mdns_sd::{ServiceDaemon, ServiceInfo};
-//! use std::collections::HashMap;
+//! use case_insensitive_hashmap::CaseInsensitiveHashMap;
 //!
 //! // Create a daemon
 //! let mdns = ServiceDaemon::new().expect("Failed to create daemon");
@@ -73,7 +73,7 @@
 //! let host_ipv4 = "192.168.1.12";
 //! let host_name = "192.168.1.12.local.";
 //! let port = 5200;
-//! let mut properties = HashMap::new();
+//! let mut properties = CaseInsensitiveHashMap::new();
 //! properties.insert("property_1".to_string(), "test".to_string());
 //! properties.insert("property_2".to_string(), "1234".to_string());
 //!
--- a/src/service_info.rs
+++ b/src/service_info.rs
@@ -1,13 +1,9 @@
 #[cfg(feature = "logging")]
 use crate::log::error;
 use crate::{dns_parser::current_time_millis, Error, Result};
+use case_insensitive_hashmap::CaseInsensitiveHashMap;
 use if_addrs::Ifv4Addr;
-use std::{
-    collections::{HashMap, HashSet},
-    convert::TryInto,
-    net::Ipv4Addr,
-    str::FromStr,
-};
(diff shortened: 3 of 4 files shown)


dalepsmith:
Not sure if this is the best way to handle this, as it changes the pubic api.

keepsimple1:
> Not sure if this is the best way to handle this, as it changes the pubic api.

It's okay to change API when it's indeed needed. I think this is a good opportunity to define our own type for the `properties` (i.e. TXT record content), something like `struct TxtProperties`?  Such a type can help us provide a stable API. The use of `CaseInsensitiveHashMap` would be an implementation detail and then we don't need to expose it as normally we want to avoid exposing types from a dependency. Does that make sense?

dalepsmith:
On 11/12/22, keepsimple1 ***@***.***> wrote:
>> Not sure if this is the best way to handle this, as it changes the pubic
>> api.
>
> It's okay to change API when it's indeed needed. I think this is a good
> opportunity to define our own type for the `properties` (i.e. TXT record
> content), something like `struct TxtProperties`?  Such a type can help us
> provide a stable API. The use of `CaseInsensitiveHashMap` would be an
> implementation detail and then we don't need to expose it as normally we
> want to avoid exposing types from a dependency. Does that make sense?

Yes, that's a very good idea!

-Dale

keepsimple1 (commented):
Also, can we use [`String::to_lowercase(&key)`](https://doc.rust-lang.org/std/string/struct.String.html#method.to_lowercase) to treat them case-insensitive? So that we don't need to add the dependency of `case_insensitive_hashmap`.

dalepsmith:
On 11/12/22, keepsimple1 ***@***.***> wrote:
> @keepsimple1 commented on this pull request.
>
> Also, can we use
> [`String::to_lowercase(&key)`](https://doc.rust-lang.org/std/string/struct.String.html#method.to_lowercase)
> to treat them case-insensitive? So that we don't need to add the dependency
> of `case_insensitive_hashmap`.

That would work I think.  The advantage of the CaseIndependentHashMap
is that it is case preserving.  Does that matter?  Will users of the
crate care if the keys are always coming back lowercased?

Personally, I'd like to see the case preserved.  My main use of this
is for discovering LXI Instruments, and they define several txt keys,
like "SerialNumber" and "FirmwareVersion".

I wonder how avahi and bonjour handle this.

-Dale

keepsimple1:
> The advantage of the CaseIndependentHashMap
is that it is case preserving.  Does that matter?  Will users of the
crate care if the keys are always coming back lowercased?

I'm not totally sure. But I think you have a point. Since `CaseIndependentHashMap` would be an implementation detail, it's not a problem. We can use it.

> I wonder how avahi and bonjour handle this.

That's a good question. Will be interesting to find out.

keepsimple1:
>  I think this is a good opportunity to define our own type for the `properties` (i.e. TXT record
 content), something like `struct TxtProperties`?  

Another idea is to define a trait `TxtProperties`, something like: 

```
trait TxtProperties {
     // The returned HashMap is keyed by original_key.lower_case() to be case insensitive.
    fn into_properties(self) -> Result<HashMap<String, TxtProperty>>;
}

struct TxtProperty {
    key: String,  // Original key that preserves the cases. We will check it to make sure only printable ASCII.
    val: Vec<u8>,  // support non-UTF8 values.
}
```

Via the trait, we can support the existing code calling with a plain `HashMap<String, String>`: 

```
impl TxtProperties for HashMap<String, String> {
<snip>
}
```

and then `ServiceInfo::new()` takes the trait for its parameter: 
```
    pub fn new<Ip: AsIpv4Addrs, P: TxtProperties>(
        ty_domain: &str,
        my_name: &str,
        host_name: &str,
        host_ipv4: Ip,
        port: u16,
        properties: Option<P>,
    ) -> Result<Self> {
```
It is basically similar with the way we use `AsIpv4Addrs` today.

keepsimple1:
@dalepsmith I got around to create a patch with some ideas I mentioned before: https://github.com/keepsimple1/mdns-sd/pull/82. Please let me know what you think, and if it can work for you. Thanks.

keepsimple1:
This PR is superseded by #82 . Closing.

Issue #94, opened 2023-03-03
Allow non-standard service names longer than 15 bytes


smith61:
I know the DNS-SD spec specifies that the limit for service names is 15 bytes (plus one for leading _), but some organizations have decided to ignore that part of the spec and use service names that are longer than 15 bytes. For example Google has decided that Android TV would use the service name _androidtvremote2 (a total of 17 bytes including the _).

I've tested with removing that specific check in "check_service_name" and did not notice any issue and clients were able to successfully resolve the service.

keepsimple1:
The limit of 15 bytes for the service name is tight. I think the more important limit is the total 255 bytes for the fully qualified service domain name. Let me check around more.

smith61:
Based on my reading of the spec, it does seem to imply that the service name limit is intended to allow for most of the 255 byte limit to go towards the instance and service/parent domain portions of the FQDN. I haven't seen any issues with clients being unable to parse these non-spec compliant FQDNs, but it potentially is possible that some clients do enforce these constraints.

I can see wanting to keep this crate to be as close to correct standards implementing as possible, but it would be nice to have this ability. Maybe hidden behind an "enable_non_standard_service_name_i_know_what_im_doing" call so people have to acknowledge that this may not work with all clients.

keepsimple1:
I checked the current code, we only call `check_service_name` during `register`. I.e. we only impose the 15-char limit when publishing a service, and we do not impose this limit when browsing the services. 

In other words, a client using this lib can currently resolve a service name that is longer than 15 bytes. I did a test and it worked. 

```
$ cargo run --example query _androidtvremote2._udp
    Finished dev [unoptimized + debuginfo] target(s) in 0.05s
     Running `target/debug/examples/query _androidtvremote2._udp`
At 144.447µs : Received other event: SearchStarted("_androidtvremote2._udp.local. on addrs [192.168.0.116]")
At 304.511134ms : Received other event: ServiceFound("_androidtvremote2._udp.local.", "pi._androidtvremote2._udp.local.")
At 304.546957ms: Resolved a new service: pi._androidtvremote2._udp.local. IP: {192.168.0.114} TXT properties: TxtProperties { properties: [TxtProperty { key: "path", val: Some([116, 119, 111]) }] }
```

Did you mean you wanted to use this lib to publish a service (i.e. running as a server) with a service name longer than 15 bytes?

smith61:
[...]

PR #96, opened 2023-03-05
Allow non-standard max length for a service name

keepsimple1:
This is to solve issue #94 .

A new method `ServiceDaemon.set_service_name_len_max()` is added to support setting a custom max length for the service name.

A public constant `SERVICE_NAME_LEN_MAX_DEFAULT` is defined for the default / standard value.

Service name length error reporting (a trade-off): it will no longer return Error directly from `register()` method for service name length errors. Instead, it will be reported in error log and `DaemonEvent::Error` received by any monitors. The register itself will not succeed in such error cases. See the test case changes for an example.

### DIFF
--- a/src/service_daemon.rs
+++ b/src/service_daemon.rs
@@ -1663,6 +1663,7 @@ impl DnsCache {
 /// The length of Service Domain name supported in this lib.
 const DOMAIN_LEN: usize = "._tcp.local.".len();
 
+/// Validate the length of "service_name" in a "_<service_name>.<domain_name>." string.
 fn check_service_name_length(ty_domain: &str, limit: u8) -> Result<()> {
     println!("check_service_name_length: {}", ty_domain);
     let service_name_len = ty_domain.len() - DOMAIN_LEN - 1; // exclude the leading `_`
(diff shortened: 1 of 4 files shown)


src/service_daemon.rs
```diff
@@ -1598,10 +1660,26 @@ impl DnsCache {
     }
 }
 
+/// The length of Service Domain name supported in this lib.
+const DOMAIN_LEN: usize = "._tcp.local.".len();
+
+/// Validate the length of "service_name" in a "_<service_name>.<domain_name>." string.
+fn check_service_name_length(ty_domain: &str, limit: u8) -> Result<()> {
+    println!("check_service_name_length: {}", ty_domain);
```
smith61, line 1668: Unneeded println.

smith61:
This PR resolves my issue. The only thing I noticed is the unneeded println in check_service_name_length

keepsimple1:
> This PR resolves my issue. The only thing I noticed is the unneeded println in check_service_name_length

Thanks for checking! I will remove the println and get it merged.

Issue #76, opened 2022-11-12
Case of keys should be ignored


dalepsmith:
From the RFC:
https://www.rfc-editor.org/rfc/rfc6763#section-6.4 Rules for Keys in DNS-SD Key/Value Pairs

Case is ignored when interpreting a key, so "papersize=A4",
"PAPERSIZE=A4", and "Papersize=A4" are all identical.

dalepsmith:
This for case #77

keepsimple1:
@dalepsmith could you please help check if  https://github.com/keepsimple1/mdns-sd/pull/82 works for you?

dalepsmith:
Yes!  I've been meaning to get to that.  Life is kind of busy at the
moment.  In the middle of a job search.

Thanks for your patience.

-Dale

dalepsmith:
On 1/26/23, keepsimple1 ***@***.***> wrote:
> @dalepsmith could you please help check if
> https://github.com/keepsimple1/mdns-sd/pull/82 works for you?

Checking it...

-Dale

dalepsmith:
So when I publish  a service with

`avahi-publish-service Testing _foo._tcp 1234 path=one Path=two PaTh=three`

I see messages with tcpdump like

`15:05:26.632340 IP deb-test.mdns > 224.0.0.251.mdns: 0*- [0q] 5/0/0
PTR Testing._foo._tcp.local., (Cache flush) TXT "path=one" "Path=two"
"PaTh=three", (Cache flush) SRV deb-test.local.:1234 0 0, (Cache
flush) AAAA fe80::a00:27ff:fea8:a09b, (Cache flush) A 192.168.1.69
(163)`

Notice that the order of the various cases of "path" is maintained.
(That's probably a bug, but it *does* stuff several different-cased
TXT values in the packet)

My test code shows this:

`15:05:11.599137 =
{"type":"add","service":"Testing","srv":"_foo._tcp","hostname":"deb-test.local.","address":"192.168.1.69","port":1234,"txt":{"Path":"two"}}`

Good news, it's only showing one "path".
Bad news, it's not the first, but the second...

-Dale

keepsimple1:
Thank you very much for checking it! I will look into it.

keepsimple1:
Thanks again for the testing. I've made the following changes: 

[...]

PR #98, opened 2023-03-11
remove duplicated keys in received TXT record

keepsimple1:
This is to fix a remaining issue reported here: https://github.com/keepsimple1/mdns-sd/issues/76#issuecomment-1464724677 .

The problem is that we are not removing duplicated keys in TXT when receiving from other responders. Also we didn't have a test case for this path.


dalepsmith:
Works great!

PR #119, opened 2023-09-01
feat(ServiceDaemon): possibility to use a specific network interface

Marcondiro:
### The addition is not a breaking change.

Hello, for my use case I need to use a specific network interface (that is also using a link local address) so I added this feature to your fantastic library.

If you think this could be useful (I do 😁) consider merging this PR or let me know if you want me to change something before merging.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "mdns-sd"
 version = "0.7.4"
-rust-version = "1.60.0"
+rust-version = "1.61.0"
 authors = ["keepsimple <keepsimple@gmail.com>"]
 edition = "2018"
 license = "Apache-2.0 OR MIT"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,7 @@
 [package]
 name = "mdns-sd"
 version = "0.7.4"
+rust-version = "1.60.0"
 authors = ["keepsimple <keepsimple@gmail.com>"]
 edition = "2018"
 license = "Apache-2.0 OR MIT"
--- a/src/service_daemon.rs
+++ b/src/service_daemon.rs
@@ -117,6 +117,8 @@ pub type Metrics = HashMap<String, i64>;
 pub struct ServiceDaemon {
     /// Sender handle of the channel to the daemon.
     sender: Sender<Command>,
+    /// name of the network interface to use (use all interfaces if None)
+    interface_filter: Option<String>,
 }
 
 impl ServiceDaemon {
@@ -125,7 +127,12 @@ impl ServiceDaemon {
     /// The daemon (re)uses the default mDNS port 5353. To keep it simple, we don't
     /// ask callers to set the port.
     pub fn new() -> Result<Self> {
-        let zc = Zeroconf::new()?;
+        Self::create_spawn_daemon(None)
+    }
+
+    fn create_spawn_daemon(interface: Option<&str>) -> Result<Self> {
+        let interface_filter =interface.map(|s| s.to_string());
+        let zc = Zeroconf::new(&interface_filter)?;
         let (sender, receiver) = bounded(100);
 
         // Spawn the daemon thread
@@ -134,7 +141,20 @@ impl ServiceDaemon {
             .spawn(move || Self::run(zc, receiver))
             .map_err(|e| e_fmt!("thread builder failed to spawn: {}", e))?;
 
-        Ok(Self { sender })
+        Ok(Self {
+            sender,
+            interface_filter,
+        })
+    }
+
+    /// Creates a new daemon using a specific `interface` and spawns a thread to run the daemon.
+    ///
+    /// The daemon will use only the network `interface` specified by the user.
+    ///
+    /// The daemon (re)uses the default mDNS port 5353. To keep it simple, we don't
+    /// ask callers to set the port.
+    pub fn on_interface(interface: &str) -> Result<Self> {
+        Self::create_spawn_daemon(Some(interface))
     }
 
     /// Starts browsing for a specific service type.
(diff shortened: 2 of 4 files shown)


src/service_daemon.rs
```diff
@@ -117,6 +117,8 @@ pub type Metrics = HashMap<String, i64>;
 pub struct ServiceDaemon {
     /// Sender handle of the channel to the daemon.
     sender: Sender<Command>,
+    /// name of the network interface to use (use all interfaces if None)
+    interface_filter: Option<String>,
```
keepsimple1, line 121: Currently we store such state only in `Zeroconf`, it will be better to keep it that way, to be consistent. 

I see you also added this state in `Zeroconf`, so I think you don't need to add it here. For the only usage I can see is in `register()` method,  you can move the `insert_ipv4addr` work into the command receiver side, i.e. in `fn register_service`.

src/service_daemon.rs
```diff
@@ -134,7 +141,20 @@ impl ServiceDaemon {
...
+        })
+    }
+
+    /// Creates a new daemon using a specific `interface` and spawns a thread to run the daemon.
+    ///
+    /// The daemon will use only the network `interface` specified by the user.
+    ///
+    /// The daemon (re)uses the default mDNS port 5353. To keep it simple, we don't
+    /// ask callers to set the port.
+    pub fn on_interface(interface: &str) -> Result<Self> {
```
keepsimple1, line 156: Can you specify or give examples about the `interface` string?

keepsimple1 (changes requested):
Thank you for your PR! Overall looks good to me. I have a couple of comments inline. A general comment:  could you please add test(s)? 

(There are minor fmt errors in GitHub actions workflow that need to fix)

Marcondiro:
I still have to add the tests 🙂

src/service_daemon.rs
```diff
@@ -1671,14 +1702,15 @@ fn call_listener(
 }
 
 /// Returns valid IPv4 interfaces in the host system.
-fn my_ipv4_interfaces() -> Vec<Ifv4Addr> {
+fn my_ipv4_interfaces(interface: Option<&str>) -> Vec<Ifv4Addr> {
```
keepsimple1, line 1705: With this change, I realized that the name `my_ipv4_interfaces` seems not very accurate because it returns only the address info.  How about renaming it to `my_ipv4_addrs` ?

keepsimple1 (commented):
Thanks for your updates.  

I also have a related thought:  a different way to support interface filter is to add a new option in `enum DaemonOption`, instead of a new constructor. But because we create `intf_socks` in `Zeroconf::new()`, it seems a bit better to filter them during `new()`.  So I think a new constructor is fine. Just sharing my thought anyway.

Cargo.toml
```diff
@@ -1,6 +1,7 @@
 [package]
 name = "mdns-sd"
 version = "0.7.4"
+rust-version = "1.60.0"
```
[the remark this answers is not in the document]
keepsimple1: And it's okay to bump up it a bit, just not too recent (for example 1.70 is too recent)

keepsimple1:
> I still have to add the tests 🙂

keepsimple1:
It was closed briefly by mistake,  Sorry. I was trying to post a comment saying "I will review again after you add the test".

Cargo.toml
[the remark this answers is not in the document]
Marcondiro: Updated to 1.61, on 1.60 tests are failing because of a dependency.

Marcondiro:
> It was closed briefly by mistake, Sorry. I was trying to post a comment saying "I will review again after you add the test".

Yes, I'm not sure when but I will add them :D

Marcondiro:
One question: why do you bind the sockets to 0.0.0.0 ?
https://github.com/keepsimple1/mdns-sd/blob/ab24121bec6a9c4f121f7444270138e6fa69f62d/src/service_daemon.rs#L527

keepsimple1:
> One question: why do you bind the sockets to 0.0.0.0 ?
> 

As I recall, if we bind the socket to a specific address, the `join_multicast_v4` and/or `set_multicast_if_v4` fails. Does this cause any issues for you?

Marcondiro:
> > One question: why do you bind the sockets to 0.0.0.0 ?
> 
> As I recall, if we bind the socket to a specific address, the `join_multicast_v4` and/or `set_multicast_if_v4` fails. Does this cause any issues for you?

Oh, ok thanks. I would prefer to generate traffic only on the interface specified

keepsimple1:
> > > One question: why do you bind the sockets to 0.0.0.0 ?
> > 
> > 
> > As I recall, if we bind the socket to a specific address, the `join_multicast_v4` and/or `set_multicast_if_v4` fails. Does this cause any issues for you?
> 
> Oh, ok thanks. I would prefer to generate traffic only on the interface specified

Although we bind to INADDR_ANY (0.0.0.0), the multicast traffic only generates on the interface specified. Btw we don't support unicast at this point. If that's not the case for you, then we shall try to fix.

keepsimple1:
Just wanted to let you know, I've posted a new release to fix issue #121 .  So you would need to update your diff to resolve conflicts. And if you encountered any problems when adding the test, please let me know.

Marcondiro:
> Just wanted to let you know, I've posted a new release to fix issue #121 . So you would need to update your diff to resolve conflicts. And if you encountered any problems when adding the test, please let me know.

Merged, thanks!
I'm sorry but I won't have time to work on this in the short/middle term :(

keepsimple1 (commented):
No problem, whenever you have time.

keepsimple1:
I have created a new PR that solves a problem and support this feature at the same time: #137 . Please take a look if you have time. Any comments are appreciated.

Marcondiro:
That's awesome, thank you! :)

PR #128, opened 2023-10-02
Remove env_logger and lower MSRV.

CosminPerRam:
Late edit: This PR was initially made to update env_logger to resolve an audit warning, doing so would have increased the MSRV to 1.63 but just for running the tests which included compiling the examples, which just using the library, isn't necessary, it was decided that we can just remove it, which also concluded that MSRV is actually 1.60, not the previously declared 1.61

[Cargo audit](https://crates.io/crates/cargo-audit) is a tool that audits your dependencies for crates with security vulnerabilities reported to the [RustSec Advisory Database](https://github.com/RustSec/advisory-db/).

Running `cargo audit` before resulted in:
```
Scanning Cargo.lock for vulnerabilities (58 crate dependencies)
Version:   0.2.14
Warning:   unsound
Title:     Potential unaligned read
Date:      2021-07-04
URL:       https://rustsec.org/advisories/RUSTSEC-2021-0145
Dependency tree:
atty 0.2.14
└── env_logger 0.9.3
    └── mdns-sd 0.8.0

warning: 1 allowed warning found
```

Updating env_logger to `0.10` fixes this and it doesn't introduce any breaking changes.


CosminPerRam:
Ah, apparently `1.63` is required, unless raising MSRV, this cannot be done.

keepsimple1:
> Ah, apparently `1.63` is required, unless raising MSRV, this cannot be done.

Sure, please raise MSRV to 1.63.  (It's surprising that `env_logger` would bring in `rustix`, luckily it's only used in tests).

CosminPerRam:
MSRV of the crate (excluding dev depends) is actually 1.60, not 1.61 (and with dev its 1.63). This has been found using [cargo-msrv](https://github.com/foresterre/cargo-msrv) (running `cargo msrv` returns 1.60 and `cargo msrv -- cargo t` returns 1.63 (minimum for env-logger)).
I've also added the [rust-version](https://doc.rust-lang.org/cargo/reference/manifest.html#the-rust-version-field) field to the Cargo.toml to restrict the msrv (which also shows up on the crates.io site nicely).

Cargo.toml
```diff
@@ -3,6 +3,7 @@ name = "mdns-sd"
 version = "0.8.0"
 authors = ["keepsimple <keepsimple@gmail.com>"]
 edition = "2018"
+rust-version = "1.60.0"
```
keepsimple1, line 6: To make thing simpler, let's use `1.63.0` here as well.

README.md
```diff
@@ -32,7 +32,7 @@ Currently this library has the following limitations:
 
 ## Minimum Rust version
 
-Tested against Rust 1.61.0
+The Minimum Supported Rust Version is 1.60.0, running tests requires 1.63.
```
keepsimple1, line 35: Like mentioned above, we can just move to 1.63.0

keepsimple1 (commented):
Thanks for your PR! Let's move the whole crate to rust 1.63, to simplify things.

keepsimple1:
Sorry, for second thought, I think it is better to remove the dev-dependency of `env_logger`.  It is actually only used in `examples/` .  I think we don't really need it and a typical Rust user can always enable logging output using their favorite logger. Does it make sense?

keepsimple1 (approved):
Looks good! Thank you for updating!

PR #130, opened 2023-10-04
Ipv6

### LINKED ISSUE #61: Ipv6 support
Might be nice to detail what is missing and what to add first. @keepsimple1 can you edit this issue with a list of tasks?

izissise:
Add ipv6 and AAAA records support

This will probably need a version bump as the interface is slightly modified

It now takes `IpAddr` and `if_addr::Interface` types in most places instead of the V4 ones
`host_ipv4` argument renamed to `ip`
`AsIpv4Addrs` renamed to `AsIpAddrs`

- [x] dns_parser: support "AAAA" address record.
- [x] responder: support register ServiceInfo with IPv4 only or IPv6 only, as well as both IPv4 and IPv6 in one register.
- [x] responder: recv and process queries on IPv6 interfaces if IPv6 address is configured.
- [x] responder: send response on IPv6 interface if IPv6 address is configured.
- [x] querier: send query on IPv6 interface if IPv6 address is configured.
- [x] querier: recv and handle answers on IPv6 interfaces if IPv6 address is configured.
- [x] Default setting: both IPv4 and IPv6 is enabled (if addr is configured). Provides API to disable IPv6 (or IPv4). i.e. allow the ServiceDaemon to be IPv6 only or IPv4 only.
- [x] Tests

The reviewer should look at the diff per commit as it is easier to read

Fix #61

### DIFF
--- a/src/service_daemon.rs
+++ b/src/service_daemon.rs
@@ -1879,10 +1879,7 @@ fn my_ip_interfaces() -> Vec<Interface> {
     if_addrs::get_if_addrs()
         .unwrap_or_default()
         .into_iter()
-        .filter_map(|i| match i.is_loopback() {
-            true => None,
-            false => Some(i),
-        })
+        .filter(|i| !i.is_loopback())
         .collect()
 }
 
--- a/src/service_daemon.rs
+++ b/src/service_daemon.rs
@@ -836,7 +836,7 @@ impl Zeroconf {
     fn add_poll_impl(
         poll_ids: &mut HashMap<usize, IpAddr>,
         poll_id_count: &mut usize,
-        ip: IpAddr
+        ip: IpAddr,
     ) -> usize {
         let key = *poll_id_count;
         *poll_id_count += 1;
--- a/src/service_daemon.rs
+++ b/src/service_daemon.rs
@@ -509,14 +509,14 @@ impl ServiceDaemon {
                         UnregisterStatus::NotFound
                     }
                     Some((_k, info)) => {
-                        for (ipv4, intf_sock) in zc.intf_socks.iter() {
+                        for (ip, intf_sock) in zc.intf_socks.iter() {
                             let packet = zc.unregister_service(&info, intf_sock);
                             // repeat for one time just in case some peers miss the message
                             if !repeating && !packet.is_empty() {
                                 let next_time = current_time_millis() + 120;
                                 zc.retransmissions.push(ReRun {
                                     next_time,
-                                    command: Command::UnregisterResend(packet, *ipv4),
+                                    command: Command::UnregisterResend(packet, *ip),
                                 });
                                 zc.timers.push(next_time);
                             }
@@ -530,10 +530,10 @@ impl ServiceDaemon {
                 }
             }
 
-            Command::UnregisterResend(packet, ipv4) => {
-                if let Some(intf_sock) = zc.intf_socks.get(&ipv4) {
+            Command::UnregisterResend(packet, ip) => {
+                if let Some(intf_sock) = zc.intf_socks.get(&ip) {
                     debug!("Send a packet length of {}", packet.len());
-                    send_packet(&packet[..], &zc.broadcast_addr, intf_sock);
+                    broadcast_on_intf(&packet[..], intf_sock);
                     zc.increase_counter(Counter::UnregisterResend, 1);
                 }
             }
@@ -590,7 +590,7 @@ fn new_socket_bind(intf: &Interface) -> Result<Socket> {
     let intf_ip = &intf.ip();
     match intf_ip {
         IpAddr::V4(ip) => {
-            let addr = SocketAddrV4::new(Ipv4Addr::new(0, 0, 0, 0).into(), MDNS_PORT);
+            let addr = SocketAddrV4::new(Ipv4Addr::new(0, 0, 0, 0), MDNS_PORT);
             let sock = new_socket(addr.into(), true)?;
 
             // Join mDNS group to receive packets.
@@ -610,7 +610,7 @@ fn new_socket_bind(intf: &Interface) -> Result<Socket> {
         }
         IpAddr::V6(ip) => {
             let mut addr = SocketAddrV6::new(
-                Ipv6Addr::new(0, 0, 0, 0, 0, 0, 0, 0).into(),
+                Ipv6Addr::new(0, 0, 0, 0, 0, 0, 0, 0),
                 MDNS_PORT,
                 0,
                 0,
@@ -694,9 +694,6 @@ struct Zeroconf {
     /// Local registered services， keyed by service full names.
     my_services: HashMap<String, ServiceInfo>,
 
-    /// Well-known mDNS IPv4 address and port
-    broadcast_addr: SockAddr,
-
     cache: DnsCache,
 
     /// Active "Browse" commands.
@@ -729,7 +726,7 @@ impl Zeroconf {
         // Get interfaces.
         let my_ifaddrs = my_ip_interfaces();
 
-        // Create a socket for every IPv4 interface.
+        // Create a socket for every IP addr.
         let mut intf_socks = HashMap::new();
         for intf in my_ifaddrs {
             let sock = match new_socket_bind(&intf) {
@@ -739,10 +736,10 @@ impl Zeroconf {
                     continue;
                 }
             };
+
             intf_socks.insert(intf.ip(), IntfSock { intf, sock });
         }
 
-        let broadcast_addr = SocketAddrV4::new(GROUP_ADDR_V4, MDNS_PORT).into();
         let monitors = Vec::new();
         let service_name_len_max = SERVICE_NAME_LEN_MAX_DEFAULT;
 
@@ -753,7 +750,6 @@ impl Zeroconf {
             poll_ids: HashMap::new(),
             id_count: 0,
             my_services: HashMap::new(),
-            broadcast_addr,
             cache: DnsCache::new(),
             queriers: HashMap::new(),
             retransmissions: Vec::new(),
@@ -994,7 +990,7 @@ impl Zeroconf {
             );
         }
 
-        self.send(&out, &self.broadcast_addr, intf_sock);
+        broadcast_dns_on_intf(&out, intf_sock);
         true
     }
 
@@ -1065,7 +1061,7 @@ impl Zeroconf {
             );
         }
 
-        self.send(&out, &self.broadcast_addr, intf_sock)
+        broadcast_dns_on_intf(&out, intf_sock)
     }
 
     /// Binds a channel `listener` to querying mDNS domain type `ty`.
@@ -1075,34 +1071,12 @@ impl Zeroconf {
         self.queriers.insert(ty, listener);
     }
 
-    /// Sends an outgoing packet, and returns the packet bytes.
-    fn send(&self, out: &DnsOutgoing, addr: &SockAddr, intf: &IntfSock) -> Vec<u8> {
-        let qtype = if out.is_query() { "query" } else { "response" };
-        debug!(
-            "Sending {} to {:?}: {} questions {} answers {} authorities {} additional",
-            qtype,
-            addr.as_socket(),
-            out.questions.len(),
-            out.answers.len(),
-            out.authorities.len(),
-            out.additionals.len()
-        );
-        let packet = out.to_packet_data();
-        if packet.len() > MAX_MSG_ABSOLUTE {
-            error!("Drop over-sized packet ({})", packet.len());
-            return Vec::new();
-        }
-
-        send_packet(&packet[..], addr, intf);
-        packet
-    }
-
     fn send_query(&self, name: &str, qtype: u16) {
         debug!("Sending multicast query for {}", name);
         let mut out = DnsOutgoing::new(FLAGS_QR_QUERY);
         out.add_question(name, qtype);
         for (_, intf_sock) in self.intf_socks.iter() {
-            self.send(&out, &self.broadcast_addr, intf_sock);
+            broadcast_dns_on_intf(&out, intf_sock);
         }
     }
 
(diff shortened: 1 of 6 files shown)


src/service_daemon.rs
```diff
@@ -582,32 +584,69 @@ impl ServiceDaemon {
...
-fn new_socket(ipv4: Ipv4Addr, port: u16, non_block: bool) -> Result<Socket> {
-    let fd = Socket::new(socket2::Domain::IPV4, socket2::Type::DGRAM, None)
+fn new_socket(addr: SocketAddr, non_block: bool) -> Result<Socket> {
+    let domain = if addr.is_ipv6() {
+        socket2::Domain::IPV6
+    } else if addr.is_ipv4() {
+        socket2::Domain::IPV4
+    } else {
+        panic!()
+    };
```
keepsimple1, line 648: Wouldn't a `match &addr` also work here, and can avoid `panic!` ?

src/service_daemon.rs
```diff
@@ -299,12 +301,12 @@ impl ServiceDaemon {
...
 
         // Add mDNS sockets to the poller.
-        for (ipv4, if_sock) in zc.intf_socks.iter() {
-            // It is OK to convert to `usize` here as we only support 32-bit
-            // or 64-bit platforms.
-            let key = u32::from(*ipv4) as usize;
+        for (ip, if_sock) in zc.intf_socks.iter() {
+            let key = zc.id_count;
+            zc.id_count += 1;
+            let _ = zc.poll_ids.insert(key, *ip);
```
keepsimple1, line 307: Nice! 

I think you might be able to go even further, i.e. combine these 3 lines into a method of `zc`:  for example:

`let key = zc.add_poll(*ip)`  

where `add_poll` does the above, and returns the `key`.

src/service_daemon.rs
```diff
@@ -740,76 +782,77 @@ impl Zeroconf {
...
+                    debug!("bind a socket to {}: {}. Skipped.", &intf.ip(), e);
                     continue;
                 }
             };
 
             // Add the new interface into the poller.
-            let key = u32::from(new_ip) as usize;
+            let key = self.id_count;
+            self.id_count += 1;
+            let _ = self.poll_ids.insert(key, new_ip);
```
keepsimple1, line 855: If you combine them into a method, then don't need to repeat the steps here, and just make a call, something like `let key = self.add_poll(new_ip)`.

src/service_daemon.rs
```diff
@@ -740,76 +782,77 @@ impl Zeroconf {
...
+        let my_ifaddrs = my_ip_interfaces();
 
         // Remove unused sockets in the poller.
         let deleted_addrs = self
             .intf_socks
             .iter()
             .filter_map(|(_, if_sock)| {
-                if !my_ifv4addrs.contains(&if_sock.intf) {
+                if !my_ifaddrs.contains(&if_sock.intf) {
                     if let Err(e) = self.poller.delete(&if_sock.sock) {
```
keepsimple1, line 813: We should also delete the entry in the new `zc.poll_ids`.  Otherwise if the host IP keeps changing, `zc.poll_ids` will grow and never shrink. 

Although deleting an entry in `HashMap` with only the value is not very efficient,  `poll_ids` is small especially if we remove stale entries promptly.

src/service_daemon.rs
```diff
@@ -638,21 +676,24 @@ struct ReRun {
...
 struct Zeroconf {
     /// Local interfaces with sockets to recv/send on these interfaces.
-    intf_socks: HashMap<Ipv4Addr, IntfSock>,
+    intf_socks: HashMap<IpAddr, IntfSock>,
+
+    /// Map poll id to IpAddr
+    poll_ids: HashMap<usize, IpAddr>,
+
+    /// Next poll id value
+    id_count: usize,
```
keepsimple1, line 692: I think naming it `poll_id_count` will be more clear.

src/service_daemon.rs
```diff
@@ -1371,23 +1400,32 @@ impl Zeroconf {
...
+                if qtype == TYPE_A || qtype == TYPE_AAAA || qtype == TYPE_ANY {
                     for service in self.my_services.values() {
                         if service.get_hostname() == question.entry.name.to_lowercase() {
                             let intf_addrs = service.get_addrs_on_intf(&intf_sock.intf);
-                            if intf_addrs.is_empty() && qtype == TYPE_A {
+                            if intf_addrs.is_empty() && (qtype == TYPE_A || qtype == TYPE_AAAA) {
+                                let t = match qtype {
+                                    TYPE_A => "TYPE_A",
+                                    TYPE_AAAA => "TYPE_AAAA",
+                                    _ => panic!(),
```
keepsimple1, line 1411: To avoid confusion, can you please do one of these two things:

1) use something str like "invalid type" to replace `panic`,

or 

2) add a function like `fn qtype_str(qtype: u16) -> &str` that implements conversion for possible qtype values (including a hard-coded &str for invalid values).  And then use the function here?

keepsimple1 (commented):
Thanks for your PR! Nice work! 

Overall looks good to me. Only have a few minor comments inline. Yes we would need a version upgrade due to the changes in API  (lesson learned: we didn't have to use IPv4 specific name/types even though we didn't support IPv6).  I will think a bit more regarding the public API change and make sure it's okay. 

Thanks again and this IPv6 support will be very helpful.

src/service_daemon.rs
izissise: I went with the first solution,  when/if qtype is an enum is will be easier to implement display on it

src/service_daemon.rs
izissise: Done, indeed better remove stall addresses

src/service_daemon.rs
izissise: I had a little fight with the borrow checker for this one, I ended with something a bit weird, maybe there is a better way.

src/service_daemon.rs
izissise: Yes, its much better with a match!

src/service_daemon.rs
```diff
@@ -355,15 +357,15 @@ impl ServiceDaemon {
                         }
 
                         // Read until no more packets available.
-                        let ipv4 = (ev.key as u32).into();
-                        while zc.handle_read(&ipv4) {}
+                        let ip = *zc.poll_ids.get(&ev.key).unwrap();
```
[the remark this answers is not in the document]
izissise: Done it logs an error now

src/service_daemon.rs
keepsimple1: Indeed the borrow checker wouldn't like the simple conversion I mentioned. Thanks for the update.

izissise:
I'm not sure why it doesn't work on windows :/

keepsimple1:
> I'm not sure why it doesn't work on windows :/

I tried it out on Windows 11.  It seems like IPv6 `IN6ADDR_ANY` does not like `set_scope_id` on Windows.  I tried the following diff, and the test passed on Windows: 

```diff
diff --git a/src/service_daemon.rs b/src/service_daemon.rs
index 08053e9..23c1e2e 100644
--- a/src/service_daemon.rs
+++ b/src/service_daemon.rs
@@ -642,9 +642,9 @@ fn new_socket_bind(intf: &Interface) -> Result<Socket> {
             Ok(sock)
         }
         IpAddr::V6(ip) => {
-            let mut addr =
+            let addr =
                 SocketAddrV6::new(Ipv6Addr::new(0, 0, 0, 0, 0, 0, 0, 0), MDNS_PORT, 0, 0);
-            addr.set_scope_id(intf.index.unwrap_or(0));
+            // addr.set_scope_id(intf.index.unwrap_or(0));
             let sock = new_socket(addr.into(), true)?;
 
             // Join mDNS group to receive packets.
```

keepsimple1:
Also I found another test failure on macOS Monterey. It seems depending on what kind of interfaces exist on the system. But anyway, the issue is that in my macOS system, there are two IPv6 interfaces with different names, but the same IPv6 address. 

I applied the following diff in the test for it to pass: 

```diff
diff --git a/tests/mdns_test.rs b/tests/mdns_test.rs
index 7350c4a..9d825e1 100644
--- a/tests/mdns_test.rs
+++ b/tests/mdns_test.rs
@@ -2,7 +2,7 @@ use if_addrs::{IfAddr, Interface};
 use mdns_sd::{
     DaemonEvent, IntoTxtProperties, ServiceDaemon, ServiceEvent, ServiceInfo, UnregisterStatus,
 };
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
 use std::net::{IpAddr, Ipv4Addr, Ipv6Addr};
 use std::sync::{Arc, Mutex};
 use std::thread::sleep;
@@ -22,7 +22,8 @@ fn integration_success() {
         .unwrap();
     let instance_name = now.as_micros().to_string(); // Create a unique name.
 
-    let my_ifaddrs: Vec<_> = my_ip_interfaces().iter().map(|intf| intf.ip()).collect();
+    let ifaddrs_set: HashSet<_> = my_ip_interfaces().iter().map(|intf| intf.ip()).collect();
+    let my_ifaddrs: Vec<_> = ifaddrs_set.into_iter().collect();
     let my_addrs_count = my_ifaddrs.len();
     println!("My IP addr(s): {:?}", &my_ifaddrs);
 ```

(Note that `ServiceDaemon` already uses a set when resolving, so didn't have two entries with the same IPv6 addrs).

**Comment by izissise** [comment]
Nice catch! 
I applied the MacOs patch.

For the scope_id, it is extremely confusing, In ipv6 multicast, the kernel needs to know on which interface the packet should be sent because all ip6 interface are required to get a link-local addr (fe80::) and the user might not want to broadcast on all ifaces.
It the end it is similar to the `-I` option in ping or other tools for ipv4.

Windows documentation seems to agree with that https://learn.microsoft.com/en-us/windows/win32/wininet/ip-version-6-support

It is confusing because for multicast addr, part of the address is called `scop` in the RFC https://datatracker.ietf.org/doc/html/rfc4291#section-2.7 to give a hint as how far the multicast should go.

Maybe windows only accept a scope_id when the socketaddr is a multicast address, because when it is not a multicast windows already knows on which interface it should send packets?

In conclusion I'll applied your patch, the socketaddr iface_id is already set twice right after with
```
sock.join_multicast_v6(&GROUP_ADDR_V6, intf.index.unwrap_or(0))
sock.set_multicast_if_v6(intf.index.unwrap_or(0))
```

File: src/service_daemon.rs [hunk 8]
```diff
@@ -134,18 +138,34 @@ impl ServiceDaemon {
...
-        let signal_sock = UdpSocket::bind("127.0.0.1:0")
+        Self::new_with_signal_addr(SocketAddrV4::new(LOOPBACK_V4, 0).into())
+    }
+
+    /// Creates a new daemon and spawns a thread to run the daemon.
+    ///
+    /// The daemon (re)uses the default mDNS port 5353. To keep it simple, we don't
+    /// ask callers to set the port.
+    /// This one uses an ipv6 local address for the signal mechanism.
+    pub fn new_signal_v6() -> Result<Self> {
```
keepsimple1, line 149: I missed this earlier.  I don't think we need this method. It's okay to just use IPv4 loopback for signaling. 

More importantly, signaling is an implementation detail. We don't want to expose it in public API.

keepsimple1 (commented):
Happy to see all tests passed! I missed one part and have a new comment. The rest looks good! I will be ready to merge.

src/service_daemon.rs
izissise: I was testing with an IPv6 only host, had an issue and though it was related but it does work using ipv4 local host.
That's a stretch but may there are OSes without an IPv4 stack implementation out there?
Anyway I'll remove the code

keepsimple1: Thanks! My understanding is that all IP stacks implements at least IPv4. If it turned out there was exceptions, we can always come back and refactor the code to support it.

keepsimple1 (approved):
Looks good, thanks! I did a quick test locally, it requires very little change of code using the API even thought it changed.

One thing I noticed is that some `Ipv4Addr` methods are not supported by `IpAddr`, for example [`is_link_local()`](https://doc.rust-lang.org/std/net/struct.Ipv4Addr.html#method.is_link_local),  which exists for `Ipv6addr` under a different name `is_unicast_link_local()`.   So a user will not be able to call `is_link_local()` on the addrs returns from `ServiceInfo.get_addresses()` .  I will open a follow-up PR to mitigate this problem.

keepsimple1:
> * Default setting: both IPv4 and IPv6 is enabled (if addr is configured). Provides API to disable IPv6 (or IPv4). i.e. allow the ServiceDaemon to be IPv6 only or IPv4 only.

I missed to ask about this point in your description. How is `ServiceDaemon` allowed to be IPv6 only or IPv4 only? (I couldn't see it now).

izissise:
It works if the interface has no IPv6 or no IPv4 configured, the ServiceDaemon won't try to broadcast on an un-configured ip version.

keepsimple1:
> It works if the interface has no IPv6 or no IPv4 configured, the ServiceDaemon won't try to broadcast on an un-configured ip version.

That's true.  I was wondering about the cases when the user wanted to explicitly disable IPv4 or IPv6 even if they are configured on the interface. It would be helpful for one of my use cases. I'll take a look.

keepsimple1:
There are two levels of this kind of control:   

1) at the `ServiceDaemon` level, effects everything in the daemon. There is an open PR to limit the daemon to a single interface, which is a concrete related case. Setting IPv4 or IPv6 only is another case. We might consider a more generic way to support these.

2) at the individual service level.  For example it probably be helpful to add a method to `ServiceInfo` to allow the user register a service only for IPv4 or IPv6, without affecting the whole daemon / other services.  (I'm thinking of implement this first.)

For browsing,  the current API is simple and does not allow much options.  But at least a user can filter out the resolved IPv4 or IPv6 addresses based their needs.

Any suggestions or comments based on your experience involving both IPv4 and IPv6?

izissise:
Actually I've been using mdns-sd on an Ipv6 only host, I agree that the user should be able to choose the interface and ip versions, but the most important thing it to have a good default behavior.

PR #140, opened 2023-10-22
fix PTR expiration preventing later service resolution

Mornix:
Hi,

I ran across this issue while using mdns-sd when browsing for a specific service domain. The bug requires the following message pattern to manifest:

1. Remote end sends an announcement (PTR+SRV+TXT) -- mdns-sd correctly emits `ServiceFound` and `ServiceResolved`
2. Remote end sends an invalidation for only the PTR (setting TTL to zero) -- mdns-sd correctly emits `ServiceRemoved`
3. Remote end sends an announcement (PTR+SRV+TXT) before any of the SRV, TXT, A, or AAAA records expire -- mdns-sd emits `ServiceFound` but does not emit `ServiceResolved`

From my understanding of the code, it looks like it assumes that a new SRV or TXT (or another non-PTR) record will be received after a new PTR relying on that to trigger the block responsible for resolving the service. 

I've added a test (in the place of least resistance) that reproduces this behaviour.

### DIFF
--- a/src/service_daemon.rs
+++ b/src/service_daemon.rs
@@ -1527,6 +1527,10 @@ impl Zeroconf {
                             name,
                             ServiceEvent::ServiceFound(name.to_string(), dns_ptr.alias.clone()),
                         );
+                        changes.push(InstanceChange {
+                            ty,
+                            name: dns_ptr.alias.clone(),
+                        });
                     }
                 } else {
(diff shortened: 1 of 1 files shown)


keepsimple1 (approved):
Thanks for your PR! That's a good fix for a subtle case!

Issue #143, opened 2023-10-28
Malformed mDNS query causes ServiceDaemon to panic


JKRhb:
First of all, thank you for this library :)

A while ago, I've been using mDNS for local DNS-SD for a small project without a problem. Revisiting the project, however, I've noticed a strange behavior of the `ServiceDaemon` which now panics immediately after startup.

After some brief investigation, I've found out that the panic is occurring in these lines in `dns_parse.rs`, with the culprit being an Amazon Firestick constantly sending malformed mDNS queries:

https://github.com/keepsimple1/mdns-sd/blob/245da0d996e1f3108b6d3813041c0979ec9df39d/src/dns_parser.rs#L1131-L1135

Would it be possible to change the error handling here to make the `ServiceDaemon` a bit more robust in the face of malformed queries?

keepsimple1:
Thank you for opening this issue, and the detailed info! That makes sense. I've opened a PR to fix this bug. If possible, please let me know if the patch would work for you.

JKRhb:
Has been resolved by https://github.com/keepsimple1/mdns-sd/pull/144 – thank you once more :)

PR #144, opened 2023-10-29
Remove un-necessary panic

keepsimple1:
This is to fix issue #143


JKRhb (approved):
Thank you for this blazingly fast fix! Works like a charm now :)

PR #159, opened 2023-12-23
Fix panic due to range out of bounds in txt record parsing

Raphiiko:
I noticed the following panic in my crash reporting earlier:

```
range end index 303 out of range for slice of length 288 (C:\...\src\service_info.rs:645:25)
```

I'm not entirely sure if this would be the correct solution, but I've made an attempt.

### DIFF
--- a/src/service_info.rs
+++ b/src/service_info.rs
@@ -642,7 +642,11 @@ fn decode_txt(txt: &[u8]) -> Vec<TxtProperty> {
         }
         offset += 1; // move over the length byte
 
-        let kv_bytes = &txt[offset..offset + length];
+        let offset_end = offset + length;
+        if offset_end > txt.len() {
+            break; // Would be out of range, skipping this property.
+        }
+        let kv_bytes = &txt[offset..offset_end];
 
         // split key and val using the first `=`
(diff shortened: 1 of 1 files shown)


src/service_info.rs
```diff
@@ -642,7 +642,11 @@ fn decode_txt(txt: &[u8]) -> Vec<TxtProperty> {
         }
         offset += 1; // move over the length byte
 
-        let kv_bytes = &txt[offset..offset + length];
+        let offset_end = offset + length;
+        if offset_end > txt.len() {
+            break; // Would be out of range, skipping this property.
```
keepsimple1, line 647: Make sense! Could you also add a log of `offset_end` and `txt.len()` when this happens?  I would say `error!()` level.  And if `error!()` doesn't work for you, `debug!()` level is also fine. 

Can you also add a unit test in this file inside `mod tests` for this negative case?  A simple test input would be fine.

keepsimple1 (commented):
Thanks for your PR! Fix looks nice, only a minor comment inline.

src/service_info.rs
Raphiiko: Gotcha, I've give given it a go!

keepsimple1 (approved):
Looks good! Thanks!

PR #164, opened 2024-01-14
netmask -> subnet

amfaber:
Hi, fantastic library I am over the moon to be rid of my C dependency on libavahi-client!
However, I've experienced some pretty strange behavior - using two wired connections, I am only able to discover services on one of them at a time. Even more strangely, which wired connection is used is random for each invocation of the program! I think I've traced the source to this change
https://github.com/keepsimple1/mdns-sd/commit/545631a051def8805fa96d053a14ae5dadc3a295

It seems the intent was to not send multiple queries to the same subnet, but the implementation instead ended up excluding any networks that use the same netmask.

As an example, my two wired connections both have the netmask 255.255.255.0, but their ip addresses and therefore subnets are completely different. Since the interfaces are stored in a hashmap, this lead to some rather strange behavior in which the order of iteration of the hashmap (which is random for each invocation of the hashmap) determined which interface was allowed to be used.

The fix seems pretty simple - get the subnet of the interface by a bitwise & of the address and the netmask and use this to check if we've already sent on that subnet.

Again, thanks for the library :)

### DIFF
--- a/src/service_daemon.rs
+++ b/src/service_daemon.rs
@@ -38,7 +38,7 @@ use crate::{
         TYPE_TXT,
     },
     error::{Error, Result},
-    service_info::{ifaddr_netmask, split_sub_domain, ServiceInfo},
+    service_info::{ifaddr_subnet, split_sub_domain, ServiceInfo},
     Receiver,
 };
 use flume::{bounded, Sender, TrySendError};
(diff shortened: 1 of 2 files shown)


dalepsmith:
I like it.  Using the subnet make much more sense than the using the netmask.

keepsimple1 (approved):
Thank you for your PR! Changes look good, nice catch!

Issue #54, opened 2022-08-09
`ServiceEvent::ServiceRemoved` doesn't work


WilliamVenner:
Hi,

I noticed that `ServiceEvent::ServiceRemoved` never fires. I tried shutting down the service daemon, I tried stop_browse on the service daemon, but the other computer never sees that my computer stopped listening/browsing/broadcasting.

Both computers are Windows.

Am I understanding correctly this event should fire when a device goes away?

keepsimple1:
Can you please try to call [unregister](https://docs.rs/mdns-sd/0.5.5/mdns_sd/struct.ServiceDaemon.html#method.unregister) ? If it didn't work, please let me know.

WilliamVenner:
Hi, that does work!

Should the event be firing when a peer exits ungracefully though?

I want to show to the user on the UI that a peer has gone away, even if ungracefully.

WilliamVenner:
While we're here... is there any way to uniquely identify a service/peer? What's stopping me advertising as a fake device or something? Am I misunderstanding the mDNS protocol, or surely is there a way to get the responder's IP address?

keepsimple1:
I don't think our current implementation detects peer going away ungracefully, even though we use defined [TTL values](https://github.com/keepsimple1/mdns-sd/blob/2a3161b494ee83f8d49eb5a8d880e0a3b62df627/src/service_info.rs#L12). I can look into this a bit. 

I'm not sure about the identity of a service/peer other than it's endpoint (IP:port), but for mDNS, there is spec for unique resource records identified by its name, rrtype, and rrclass. A resource record can be unique or shared. See [this RFC section](https://www.rfc-editor.org/rfc/rfc6762#section-6.6). Currently we haven't implemented the Conflict Resolution mentioned in the RFC yet.

WilliamVenner:
> I don't think our current implementation detects peer going away ungracefully, even though we use defined [TTL values](https://github.com/keepsimple1/mdns-sd/blob/2a3161b494ee83f8d49eb5a8d880e0a3b62df627/src/service_info.rs#L12). I can look into this a bit.

[...]

PR #176, opened 2024-02-28
SRV TTL triggers timeout event

wfeii1980:
[issue 54](https://github.com/keepsimple1/mdns-sd/issues/54) Add ServiceEvent::ServiceInactive

### DIFF
--- a/src/service_daemon.rs
+++ b/src/service_daemon.rs
@@ -513,16 +513,31 @@ impl ServiceDaemon {
             // check and evict expired records in our cache
             let now = current_time_millis();
             let map = zc.queriers.clone();
-            zc.cache.evict_expired(now, |expired| {
-                if let Some(dns_ptr) = expired.any().downcast_ref::<DnsPointer>() {
-                    let ty_domain = dns_ptr.get_name();
-                    call_listener(
-                        &map,
-                        ty_domain,
-                        ServiceEvent::ServiceRemoved(ty_domain.to_string(), dns_ptr.alias.clone()),
-                    );
-                }
-            });
+            for (ty_domain, _sender) in zc.queriers.iter() {
+                zc.cache.evict_expired(ty_domain, now, |expired| {
+                    if let Some(dns_ptr) = expired.any().downcast_ref::<DnsPointer>() {
+                        // let ty_domain = dns_ptr.get_name();
+                        call_listener(
+                            &map,
+                            ty_domain,
+                            ServiceEvent::ServiceRemoved(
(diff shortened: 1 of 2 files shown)


keepsimple1:
Thanks for your PR! Could you please describe the problem at hand?  

From the title, my understanding is that SRV TTL expired while PTR record was still valid because PTR has 75 mins TTL while SRV only has 2 mins TTL.   Is that the problem? 

If yes, I think the direct cause is that currently we checks [only PTR record](https://github.com/keepsimple1/mdns-sd/blob/4d719a3a47152b634a0314bfd9041690772b6e29/src/service_daemon.rs#L2126) for refresh due.   If we send out `TYPE_SRV` refresh query when SRV is due to refresh, will that solve your problem?

At the high level, I think if SRV TTL indeed expires (even with refresh), which means the server no longer exists for the service instance, we should generate `ServiceRemoved` as the service instance is no longer valid.  It seems to me `ServiceInactive` is too similar to `ServiceRemoved` and would cause confusions.

wfeii1980:
Because I don't quite understand the difference between `PTR` and `SRV` timeout, I added a `ServiceInactive`.

keepsimple1:
> Because I don't quite understand the difference between `PTR` and `SRV` timeout, I added a `ServiceInactive`.

Both `PTR` and `SRV` (and address records `A` / `AAAA`) are required to have a [complete service instance](https://github.com/keepsimple1/mdns-sd/blob/42c94bbb66df55b876da35684718f119d126f2d8/src/service_info.rs#L262). I think any of them timeout should generate the event `ServiceRemoved`.  Do you need to know which of them caused the `ServiceRemoved` ?  If yes, then we need to find a way to distinguish them.

keepsimple1:
I added a draft PR #180 to address the issues here. Let me know what you think. @wfeii1980 . Thanks.

keepsimple1:
I've updated the PR #180 to support your use case.  Would you please help check if that PR can: 

1. refresh SRV when the responder is alive.
2. trigger `ServiceRemoved` when the SRV expires and the responder no longer alive

Please let me know if any issues. Thanks!

keepsimple1:
@wfeii1980 please let me know what do you think about PR #180 ?  Would it help your use case?

wfeii1980:
very good!

keepsimple1:
I've merged PR #180, which should have fixed this one. If you still have issues with SRV TTL triggers, please open a new issue, or re-open this one. Thanks!

PR #180, opened 2024-03-04
refresh SRV records and send out ServiceRemoved for expired SRV

keepsimple1:
This patch is to fix issue raised in #176.  There are two parts:

- [x] Add checks of refresh for SRV record so that they will be updated if the responder is alive.
- [x] Also triggers `ServiceRemoved` when SRV record expires as the responder is no longer active.

Added a new setter `_set_host_ttl()` for `DnsRecord` to support testing.

Also addressed issue #203 : only apply cache-flush bit if the record is new, i.e. rdata changed. Otherwise, it is only a refresh of TTL. This makes sense to me as it can avoid unnecessary new ServiceResolved events.

Also removed `-Wclippy::all` from CI as it refers the default warnings, hence is redundant with `-D warnings`.  (https://doc.rust-lang.org/clippy/)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -27,6 +27,6 @@ jobs:
       - name: Build
         run: cargo build
       - name: Run clippy and fail if any warnings
-        run: cargo clippy -- -W clippy::all -D warnings
+        run: cargo clippy -- -D warnings
       - name: Run tests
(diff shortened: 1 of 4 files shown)


keepsimple1:
Rebased to the latest `main` branch and update the PR description.

PR #185, opened 2024-03-19
query TYPE_A and TYPE_AAAA via Command::Resolve

### LINKED ISSUE #182: Services published using python-zerconf or systemd-resolved are not resolved
Example python code to publish a service:
```python
from zeroconf import Zeroconf, ServiceInfo
from socket import gethostname

zeroconf = Zeroconf()
service_type = "_workstation._tcp.local."

zeroconf.register_service(
    ServiceInfo(
        service_type,
        f"worky-station.{service_type}",
        4848,
        server=f"{gethostname()}.local",
    )
)

try:
    input("Press enter to exit...\n\n")
finally:
    zeroconf.close()
````

Running `avahi-browse -tpr _workstation._tcp` on the same Linux machine yields resolved results
```
+;eno33554984;IPv4;worky-station;Workstation;local
+;eno16777736;IPv4;worky-station;Workstation;local
+;lo;IPv4;worky-station;Workstation;local
=;eno33554984;IPv4;worky-station;Workstation;local;void-vm.local;192.168.178.76;4848;
=;eno16777736;IPv4;worky-station;Workstation;local;void-vm.local;192.168.73.130;4848;
=;lo;IPv4;worky-station;Workstation;local;void-vm.local;127.0.0.1;4848;
```

Example program i used to test resolving using mdns-sd
```rust
use mdns_sd::{ServiceDaemon, ServiceEvent};

fn main() {
    let mdns = ServiceDaemon::new().expect("Failed to create daemon");
    let receiver = mdns.browse("_workstation._tcp.local.").expect("Failed to browse");
    let mut search_done = false;
    while let Ok(event) = receiver.recv() {
        match event {
            ServiceEvent::ServiceResolved(info) => {
                println!(
                    "Resolved a new service: {} host: {} port: {} IP: {:?} TXT properties: {:?}",
                    info.get_fullname(),
                    info.get_hostname(),
                    info.get_port(),
                    info.get_addresses(),
                    info.get_properties(),
                );
            }
            ServiceEvent::SearchStarted(_service) => {
                if search_done {
                    mdns.stop_browse(srv).expect("To stop browsing");
                }
                search_done = true;
            }
            ServiceEvent::SearchStopped(_service) => {
                break;
            }
            _ => {}
        }
    }
    mdns.shutdown().unwrap();
}
```

What I also noticed, when I use `avahi-publish-service worky-station _workstation._tcp. 4848`, it can be resolved successfully using the obove rust program example.

hrzlgnm:
This fixes issues with resolving records published by python-zeroconf or systemd-resolved without an actual A or AAAA address assigned.

Closes #182 

Manually tested on linux and windows successfully

### DIFF
--- a/src/service_daemon.rs
+++ b/src/service_daemon.rs
@@ -1571,6 +1571,12 @@ impl Zeroconf {
             }
         }
 
+        // if we have no A or AAAA records yet, try to query them if we have a hostname
+        if !info.get_hostname().is_empty() && info.get_addresses().is_empty() {
+            self.send_query(info.get_hostname(), TYPE_A);
+            self.send_query(info.get_hostname(), TYPE_AAAA);
+        }
+
         Ok(info)
     }
 
--- a/src/service_daemon.rs
+++ b/src/service_daemon.rs
@@ -1580,12 +1580,6 @@ impl Zeroconf {
             }
         }
 
-        // if we have no A or AAAA records yet, try to query them if we have a hostname
-        if !info.get_hostname().is_empty() && info.get_addresses().is_empty() {
-            self.send_query(info.get_hostname(), TYPE_A);
-            self.send_query(info.get_hostname(), TYPE_AAAA);
-        }
-
         Ok(info)
     }
 


keepsimple1 (commented):
Thanks for coming up with the fix!  Good work! I have two comments: 

1. Should we use the existing `Command::Resolve` to send this query? The reason is that it is possible that the responses came in more than 1 `DnsIncoming` datagram. This will give it a chance to process completely.  It will also handle retry. 

I opened PR #186 to show what I meant. 

2.  As an optimization, we can refactor `send_query` to accept more than 1 type so that we can ask more than 1 questions in the packet. So we can reduce the network traffic. (We  / I can do this later, just a comment here)

hrzlgnm:
> Thanks for coming up with the fix! Good work! I have two comments:
> 
>     1. Should we use the existing `Command::Resolve` to send this query? The reason is that it is possible that the responses came in more than 1 `DnsIncoming` datagram. This will give it a chance to process completely.  It will also handle retry.
> 
> 
> I opened PR #186 to show what I meant.
> 
>     2. As an optimization, we can refactor `send_query` to accept more than 1 type so that we can ask more than 1 questions in the packet. So we can reduce the network traffic. (We  / I can do this later, just a comment here)

Regarding 1. yes that makes more sense to me as we wouldn't retry endlessly if we never receive a reply. 

Regarding 2: yes makes sense to send multiple queries in one packet to reduce network traffic.

keepsimple1:
> Regarding 1. yes that makes more sense to me as we wouldn't retry endlessly if we never receive a reply.
> 
> Regarding 2: yes makes sense to send multiple queries in one packet to reduce network traffic.

I think at least two options we can do: 

a) Merge your current changes in this PR and then I will follow up with changes using the other PR.  And will need your help to verify your use case still works.

b) You can update this PR with suggestion changes (particularly 1) and we merge it. 

I'm good with either approaches. Let me know what you think?

hrzlgnm:
I'd rather do b to avoid the unclean intermediate state

hrzlgnm:
I'll update my pr after work

hrzlgnm:
Update the PR and retested my 2 cases successfully on windows and linux.

hrzlgnm:
And thank you very much for the explanatory PR, it was very educational :+1:

keepsimple1 (approved):
Thank you for the update! LGTM!

Issue #65, opened 2022-09-08
Unicast Send/Receive Support


pixsperdavid:
I need unicast send/receive support for my current project so willing to work on adding this feature. Similar to #61, could you give an overview of tasks that are involved in implementing this?

keepsimple1:
Thanks for stepping up and take the initiative! Here is a list of tasks on my mind:

- [ ] sockets for receiving: currently `Zeroconf::intf_socks` only processes multicast packets, these socks could also process incoming unicast packets received.  To me, this would need some experiments as all `intf_socks` are bound to `0.0.0.0`, hence they are probably load-balanced by the kernel for unicast.  How can we tell the unicast packets from the multicast packets? 

another option is to use a different server port (i.e. not 5353) just for unicast. It will be worth to check out what existing unicast DNS-SD software do. 

- [ ] socket for sending: this is probably easier, just adding a new socket in `Zeroconf` to send all unicast outgoing packets. The bind port can be chosen by the system. 

- [ ] handle unicast domains in `browse`:  for unicast, the domain will be different from `.local.` , and similar to a regular domain like `cs.mycompany.com.`.  When detected non-local domain, we should use unicast socket to send out the query. We might also need to do a regular DNS lookup for the domain IP address.

The above is the main points I can think of for now. There are muddy areas in unicast DNS-SD I don't know enough about. Hope we can figure everything out along the way.

pixsperdavid:
Thanks for the information, I'll make a start and submit a draft PR for review when ready.

pixsperdavid:
> distinguish unicast packets from multicast packet: probably not required, as long as we process them.

Unfortunately, according to [RFC 6702 5.5](https://www.rfc-editor.org/rfc/rfc6762#section-5.5), this is required. The library should respond to any queries sent to the server's unicast address with unicast responses. Therefore the library needs to determine the destination address of the incoming packets. I did some research and there's not currently an easy way to get this information in a cross-platform way in Rust. 

The best example I've found so far is the [multicast socket](https://github.com/bltavares/multicast-socket) crate. So a potential way forward could be to base the implementation on this, although it would require separate implementations for windows and unix sockets, and would also require some unsafe code. Thoughts?

keepsimple1:
[...]

PR #187, opened 2024-03-20
prototype: add unicast sockets

keepsimple1:
This is to experiment some ideas in issue #65 .  It does not fully address the issue (details below) but showing it's possible to bind to multicast address and use separate unicast sockets. (only tested on macOS at the moment). 

In theory, it is mainly based on two statements in the book "Unix network programming" by W. Richard Stevens: 

1. In section 19.3 Multicasting versus Broadcasting on A LAN, it says:

`... Nothing special is required to send a multicast datagram: the application does not have to join the multicast group.`

hence, in this patch, we use `uni_sock` in the `fn send_packet` even for sending multicast packets.

2. In section 19.5 Multicast Socket Options, towards to the end, it says:

` .. Some applications also bind the multicast address to the socket, in addition to the port. This prevents any other datagrams that might be received for that port from being delivered to the socket.`

Hence, in this patch, we bind the multicast socket to `GROUP_ADDR_V4/6`, instead of `INADDR_ANY`, to prevent this socket from receiving unicast packets.

However, there is a problem:  even with dedicated unicast sockets, we don't know the source IP of unicast packets received. This is because we are using `.read()` , not `.recv_from()`.   This is an old problem because in `socket2` `.recv_from()` became unsafe to use since 0.4.0. 

In my understanding, a prime use case of issue #65 is that:  
1. a querier sends a query datagram to mDNS Multicast address with `QU` flag in its question.
2. a responder receives the query on its multicast socket. 
3. Due to `QU` flag, the responder sends back responses in unicast back to the querier. 
4. The querier receives the unicast responses.

Although this patch can support receiving the unicast responses, the responder cannot do step 3 as they don't know the unicast IP of the querier.

A more rare use case is that the querier directly sends questions in unicast, which could be supported by this patch.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -29,4 +29,4 @@ jobs:
       - name: Run clippy and fail if any warnings
         run: cargo clippy -- -D warnings
       - name: Run tests
-        run: RUST_LOG=debug cargo test
+        run: cargo test
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -29,4 +29,4 @@ jobs:
       - name: Run clippy and fail if any warnings
         run: cargo clippy -- -D warnings
       - name: Run tests
-        run: cargo test
+        run: RUST_LOG=debug cargo test
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,3 +25,4 @@ socket2 = { version = "0.5.5", features = ["all"] }      # socket APIs
 
 [dev-dependencies]
 fastrand = "1.8"
+env_logger = { version = "0.10", default-features = false }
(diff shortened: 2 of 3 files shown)


keepsimple1:
The test failure on Ubuntu Linux is: 

`thread 'service_daemon::tests::service_with_temporarily_invalidated_ptr' panicked at 'called `Result::unwrap()` on an `Err` value: Msg("socket bind to [ff02::fb]:5353 failed: Invalid argument (os error 22)")', src/service_daemon.rs:2496:47`

The IPv6 addr `ff02::fb` is a link-local multicast address. Binding to it works fine on macOS but for some reason it failed on Linux.

PR #190, opened 2024-04-03
Implemented unicast responses to direct unicast, unicast questions and legacy unicast queries.

pixsperdavid:
Draft for now as I now need to construct some tests for this to verify it functions as expected.

One missing area for now is the following from RFC 6762:

> When receiving a question with the unicast-response bit set, a
   responder SHOULD usually respond with a unicast packet directed back
   to the querier.  However, if the responder has not multicast that
   record recently (within one quarter of its TTL), then the responder
   SHOULD instead multicast the response so as to keep all the peer
   caches up to date, and to permit passive conflict detection.  In the
   case of answering a probe question ([Section 8.1](https://www.rfc-editor.org/rfc/rfc6762#section-8.1)) with the unicast-
   response bit set, the responder should always generate the requested
   unicast response, but it may also send a multicast announcement if
   the time since the last multicast announcement of that record is more
   than a quarter of its TTL.

Implementing the above will require storing the time in which each service record was last multicasted on each specific interface so is therefore quite complex. For now I have simply used the logic that if all the questions in a given query request a unicast response, a unicast response is sent.

### DIFF
--- a/src/service_daemon.rs
+++ b/src/service_daemon.rs
@@ -42,14 +42,15 @@ use crate::{
     Receiver,
 };
 use flume::{bounded, Sender, TrySendError};
-use if_addrs::Interface;
+use if_addrs::{IfAddr, Interface};
 use polling::Poller;
-use socket2::{SockAddr, Socket};
+use socket2::SockAddr;
+use socket_pktinfo::PktInfoUdpSocket;
+use std::cmp::min;
 use std::{
     cmp,
     collections::{HashMap, HashSet},
     fmt,
-    io::Read,
     net::{IpAddr, Ipv4Addr, Ipv6Addr, SocketAddr, SocketAddrV4, SocketAddrV6, UdpSocket},
     str, thread,
     time::Duration,
@@ -74,6 +75,8 @@ const GROUP_ADDR_V6: Ipv6Addr = Ipv6Addr::new(0xff02, 0, 0, 0, 0, 0, 0, 0xfb);
 const LOOPBACK_V4: Ipv4Addr = Ipv4Addr::new(127, 0, 0, 1);
 
 const RESOLVE_WAIT_IN_MILLIS: u64 = 500;
+
+const LEGACY_UNICAST_RR_TTL: u32 = 10;
 
 /// Response status code for the service `unregister` call.
 #[derive(Debug)]
@@ -105,6 +108,8 @@ enum Counter {
     UnregisterResend,
     Browse,
     Respond,
+    RespondUnicast,
+    RespondLegacyUnicast,
     CacheRefreshQuery,
 }
 
@@ -117,6 +122,8 @@ impl fmt::Display for Counter {
             Counter::UnregisterResend => write!(f, "unregister-resend"),
             Counter::Browse => write!(f, "browse"),
             Counter::Respond => write!(f, "respond"),
+            Counter::RespondUnicast => write!(f, "respond-unicast"),
+            Counter::RespondLegacyUnicast => write!(f, "respond-legacy-unicast"),
             Counter::CacheRefreshQuery => write!(f, "cache-refresh"),
         }
     }
@@ -691,7 +698,7 @@ impl ServiceDaemon {
 }
 
 /// Creates a new UDP socket that uses `intf` to send and recv multicast.
-fn new_socket_bind(intf: &Interface) -> Result<Socket> {
+fn new_socket_bind(intf: &Interface) -> Result<PktInfoUdpSocket> {
     // Use the same socket for receiving and sending multicast packets.
     // Such socket has to bind to INADDR_ANY or IN6ADDR_ANY.
     let intf_ip = &intf.ip();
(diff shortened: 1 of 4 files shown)


src/service_daemon.rs
```diff
@@ -1863,9 +1938,18 @@ impl Zeroconf {
...
             out.id = msg.id;
-            broadcast_dns_on_intf(&out, intf_sock);
-
-            self.increase_counter(Counter::Respond, 1);
+            if is_unicast_reply {
+                if is_legacy_unicast {
+                    unicast_dns_on_intf(&out, src_addr.ip(), intf_sock);
+                    self.increase_counter(Counter::RespondLegacyUnicast, 1);
+                } else {
+                    unicast_dns_on_intf(&out, src_addr.ip(), intf_sock);
```
keepsimple1, line 1946: nit pick:  this common call of `unicast_dns_on_intf` can be extracted out and moved to before `if is_legacy_unicast` check.

src/service_daemon.rs
```diff
@@ -1430,12 +1437,48 @@ impl Zeroconf {
...
             Ok(msg) => {
                 if msg.is_query() {
-                    self.handle_query(msg, ip);
+                    self.handle_query(msg, ip, pktinfo.addr_src, is_unicast);
                 } else if msg.is_response() {
-                    self.handle_response(msg);
+                    if !is_unicast {
+                        self.handle_response(msg);
+                    } else {
+                        error!("Invalid message: unrequested unicast response");
```
keepsimple1, line 1480: I don't fully understand this part. Does it mean we will not handle _any_ unicast responses?   

Is it because this lib does not have ability to send out query with `UNIQUE_RESPONSE` bit set yet?

keepsimple1 (commented):
Thank you for the PR! While you're working on adding the tests, here are a few comments inline.

keepsimple1:
> Draft for now as I now need to construct some tests for this to verify it functions as expected.
> 
> One missing area for now is the following from RFC 6762:
> 
> > When receiving a question with the unicast-response bit set, a
> > responder SHOULD usually respond with a unicast packet directed back
> > to the querier.  However, if the responder has not multicast that
> > record recently (within one quarter of its TTL), then the responder
> > SHOULD instead multicast the response so as to keep all the peer
> > caches up to date, and to permit passive conflict detection.  In the
> > case of answering a probe question ([Section 8.1](https://www.rfc-editor.org/rfc/rfc6762#section-8.1)) with the unicast-
> > response bit set, the responder should always generate the requested
> > unicast response, but it may also send a multicast announcement if
> > the time since the last multicast announcement of that record is more
> > than a quarter of its TTL.
> 
> Implementing the above will require storing the time in which each service record was last multicasted on each specific interface so is therefore quite complex. For now I have simply used the logic that if all the questions in a given query request a unicast response, a unicast response is sent.

Sounds good to me, to implement this part later / separately.

src/service_daemon.rs
pixsperdavid: Yes exactly. Relevant RFC passage:

> A Multicast DNS querier MUST only accept unicast responses if
   they answer a recently sent query (e.g., sent within the last two
   seconds) that explicitly requested unicast responses.  A Multicast
   DNS querier MUST silently ignore all other unicast responses.

Therefore as the library currently will never send any queries requesting a unicast response, any unicast responses received will not be in answer to queries made by the library and therefore should be ignored. I'll add a comment to indicate this.

src/service_daemon.rs
```diff
@@ -1430,12 +1437,48 @@ impl Zeroconf {
 
         buf.truncate(sz); // reduce potential processing errors
 
+        let is_unicast = !pktinfo.addr_dst.is_multicast();
+
+        // Ignore unicast packets outside the local link
+        if is_unicast {
+            let should_respond = match (pktinfo.addr_src.ip(), &intf_sock.intf.addr) {
```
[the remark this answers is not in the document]
pixsperdavid: This actually has slightly different behaviour for IPv6 than my check. RFC states:

> Since it is possible
   for a unicast query to be received from a machine outside the local
   link, responders SHOULD check that the source address in the query
   packet matches the local subnet for that link (or, in the case of
   IPv6, the source address has an on-link prefix) and silently ignore
   the packet if not.

For IPv6, the `valid_ip_on_intf` function checks the subnet mask against the IP address, whereas I check if the address has an on-link prefix.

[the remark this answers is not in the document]
keepsimple1: You are right. Never mind. (I thought it covers the ipv6 case, but it's not)

keepsimple1 (commented):
Thanks for making the changes. Looks good!

keepsimple1:
Just wanted to share that I am looking at if we can use `PktInfoUdpSocket` to solve another issue #372 .

keepsimple1:
Just to circle back, `socket-pktinfo` crate is now introduced in `main` branch via PR #373. It would save some diff from this PR.

PR #192, opened 2024-04-09
Add support for resolving non-service hostnames

### LINKED ISSUE #191: Utility function to resolve mDNS hostnames
Hello, this is a feature request, as well as a question about the project scope.

We're using this package for making a remoting cli tool, and we'd like to persist a mDNS hostname, and re-resolve it on invocation. Since this project has its own stack for all mDNS operations, the code to do this seems to already be there (`Command::Resolve`)? It would be nice if some of this functionality could be exposed to the user, either as utility functions on the `ServiceDaemon` or by making the `Zeroconf` impl into a public lower level interface.

I had some worries while making this issue about the scope of the project: the readme seems to indicate that the focus is only set on service discovery. Publicizing some of the "mDNS-only" functionality may be a slippery slope towards being an mDNS library with the service discovery parts on top. But maybe that's not the case. Or maybe that's even okay with you. What do you think?

I'd be happy to create a PR with `ServiceDaemon::resolve_service_host(&self, service_instance_fullname: String, try_count: u16)` if I get the green light.

oysteintveit-nordicsemi:
Resolves #191.

This adds several pieces of public API, for resolving mDNS hostnames as standalone.

- `ServiceDaemon::resolve(&self, hostname: &str) -> Result<Receiver<ResolutionEvent>>`
- `ServiceDaemon::stop_resolve(&self, hostname: &str) -> Result<()>`

as well as

```rust
pub enum ResolutionEvent {
    /// Started searching for the ip address of a hostname.
    SearchStarted(String),
    /// One or more addresses for a hostname has been found.
    HostnameAddressesFound(String, HashSet<IpAddr>),
    /// One or more addresses for a hostname has been removed.
    HostnameAddressesRemoved(String, HashSet<IpAddr>),
    /// Stopped searching for the ip address of a hostname.
    SearchStopped(String),
}
```

The internal implementation uses the dns cache for addresses, which *should* make it possible for service discovery and hostname discovery to benefit from each other in the case that you are running both at the same time.

The rest of the implementation is largely either copy paste of- or merging with the code that handles the service discovery, with some exceptions where the events are reporting an aggregate of ip-addresses rather than reporting one at a time.

### Other notes

I refactored `DnsCache::evict_expired` to resolve some borrow checker issues. The current implementation will evict all expired records before it starts sending out events. It should not have any effect on any users of `ServiceDaemon::browse`, since they don't have direct access to the cache.

The current implementation is a bit loud, because it's missing [Known Answer Suppression (RFC6762, sec. 7.1)](https://datatracker.ietf.org/doc/html/rfc6762#section-7.1) (which is stated as a MUST requirement). As far as I understand, the current implementation for service browsing does not have this either, so I think it should be considered as a new issue.

### DIFF
--- a/src/service_daemon.rs
+++ b/src/service_daemon.rs
@@ -246,7 +246,7 @@ impl ServiceDaemon {
         &self,
         hostname: &str,
         timeout: Option<u64>,
-    ) -> Result<Receiver<ResolutionEvent>> {
+    ) -> Result<Receiver<HostnameResolutionEvent>> {
         check_hostname(hostname)?;
         let (resp_s, resp_r) = bounded(10);
         self.send_cmd(Command::ResolveHostname(
@@ -519,25 +519,24 @@ impl ServiceDaemon {
 
             // Remove hostname resolvers with expired timeouts.
             for hostname in zc
-                .hostname_resolver_timeouts
+                .hostname_resolvers
                 .clone()
                 .into_iter()
-                .filter(|(_, t)| now >= *t)
-                .map(|(h, _)| h)
+                .filter(|(_, (_, timeout))| timeout.is_some_and(|t| now >= t))
+                .map(|(hostname, _)| hostname)
             {
                 log::debug!("hostname resolver timeout for {}", &hostname);
                 call_hostname_resolution_listener(
                     &zc.hostname_resolvers,
                     &hostname,
-                    ResolutionEvent::SearchTimeout(hostname.to_owned()),
+                    HostnameResolutionEvent::SearchTimeout(hostname.to_owned()),
                 );
                 call_hostname_resolution_listener(
                     &zc.hostname_resolvers,
                     &hostname,
-                    ResolutionEvent::SearchStopped(hostname.to_owned()),
+                    HostnameResolutionEvent::SearchStopped(hostname.to_owned()),
                 );
                 zc.hostname_resolvers.remove(&hostname);
-                zc.hostname_resolver_timeouts.remove(&hostname);
             }
 
             // process commands from the command channel
@@ -611,7 +610,7 @@ impl ServiceDaemon {
                     call_hostname_resolution_listener(
                         &zc.hostname_resolvers,
                         hostname,
-                        ResolutionEvent::HostnameAddressesRemoved(
+                        HostnameResolutionEvent::HostnameAddressesRemoved(
                             hostname.to_string(),
                             zc.cache.get_addresses_for_host(hostname),
                         ),
@@ -661,7 +660,7 @@ impl ServiceDaemon {
 
             Command::ResolveHostname(hostname, next_delay, listener, timeout) => {
                 let addr_list: Vec<_> = zc.intf_socks.keys().collect();
-                if let Err(e) = listener.send(ResolutionEvent::SearchStarted(format!(
+                if let Err(e) = listener.send(HostnameResolutionEvent::SearchStarted(format!(
                     "{} on addrs {:?}",
                     &hostname, &addr_list
                 ))) {
@@ -677,8 +676,7 @@ impl ServiceDaemon {
                     zc.query_cache_for_hostname(&hostname, listener.clone());
                 }
 
-                zc.send_query(&hostname, TYPE_A);
-                zc.send_query(&hostname, TYPE_AAAA);
+                zc.send_query_vec(&[(&hostname, TYPE_A), (&hostname, TYPE_AAAA)]);
                 zc.increase_counter(Counter::ResolveHostname, 1);
 
                 let now = current_time_millis();
@@ -688,14 +686,15 @@ impl ServiceDaemon {
 
                 // Only add retransmission if it does not exceed the hostname resolver timeout, if any.
                 if zc
-                    .hostname_resolver_timeouts
+                    .hostname_resolvers
                     .get(&hostname)
-                    .map(|timeout| next_time < *timeout)
+                    .and_then(|(_sender, timeout)| *timeout)
+                    .map(|timeout| next_time < timeout)
                     .unwrap_or(true)
                 {
                     zc.add_retransmission(
                         next_time,
-                        Command::ResolveHostname(hostname, delay, listener, timeout),
+                        Command::ResolveHostname(hostname, delay, listener, None),
                     );
                 }
             }
@@ -786,7 +785,9 @@ impl ServiceDaemon {
             },
 
             Command::StopResolveHostname(hostname) => {
-                if let Some((host, sender)) = zc.hostname_resolvers.remove_entry(&hostname) {
+                if let Some((host, (sender, _timeout))) =
+                    zc.hostname_resolvers.remove_entry(&hostname)
+                {
                     // Remove pending resolve commands in the reruns.
                     debug!("StopResolve: removed queryer for {}", &host);
                     let mut i = 0;
@@ -801,11 +802,8 @@ impl ServiceDaemon {
                         i += 1;
                     }
 
-                    // Remove the timeout, if any.
-                    zc.hostname_resolver_timeouts.remove(&hostname);
-
                     // Notify the client.
-                    match sender.send(ResolutionEvent::SearchStopped(hostname)) {
+                    match sender.send(HostnameResolutionEvent::SearchStopped(hostname)) {
                         Ok(()) => debug!("Sent SearchStopped to the listener"),
                         Err(e) => warn!("Failed to send SearchStopped: {}", e),
                     }
@@ -1049,13 +1047,9 @@ struct Zeroconf {
     service_queriers: HashMap<String, Sender<ServiceEvent>>, // <ty_domain, channel::sender>
 
     /// Active "ResolveHostname" commands.
-    hostname_resolvers: HashMap<String, Sender<ResolutionEvent>>, // <hostname, channel::sender>
-
-    /// Timeout timestamps for active "ResolveHostname" commands.
     ///
-    /// The timestamps are set at the future time when the command should timeout.
-    /// If a hostname resolver does not have a timeout, it will not be added to this map.
-    hostname_resolver_timeouts: HashMap<String, u64>, // <hostname, UNIX timestamp in millis>
+    /// The timestamps are set at the future timestamp when the command should timeout.
+    hostname_resolvers: HashMap<String, (Sender<HostnameResolutionEvent>, Option<u64>)>, // <hostname, (channel::sender, UNIX timestamp in millis)>
 
     /// All repeating transmissions.
     retransmissions: Vec<ReRun>,
@@ -1127,7 +1121,7 @@ impl Zeroconf {
             my_services: HashMap::new(),
             cache: DnsCache::new(),
             hostname_resolvers: HashMap::new(),
-            hostname_resolver_timeouts: HashMap::new(),
+            // hostname_resolver_timeouts: HashMap::new(),
             service_queriers: HashMap::new(),
             retransmissions: Vec::new(),
             counters: HashMap::new(),
@@ -1544,15 +1538,14 @@ impl Zeroconf {
     fn add_hostname_resolver(
         &mut self,
         hostname: String,
-        listener: Sender<ResolutionEvent>,
+        listener: Sender<HostnameResolutionEvent>,
         timeout: Option<u64>,
     ) {
+        let real_timeout = timeout.map(|t| current_time_millis() + t);
         self.hostname_resolvers
-            .insert(hostname.to_owned(), listener);
-        if let Some(t) = timeout {
-            let now = current_time_millis();
-            self.hostname_resolver_timeouts.insert(hostname, now + t);
-            self.timers.push(now + t);
+            .insert(hostname, (listener, real_timeout));
+        if let Some(t) = real_timeout {
+            self.timers.push(t);
         }
     }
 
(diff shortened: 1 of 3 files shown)


src/service_daemon.rs
```diff
@@ -233,6 +235,26 @@ impl ServiceDaemon {
         self.send_cmd(Command::StopBrowse(ty_domain.to_string()))
     }
 
+    /// Starts querying for the ip addresses of a hostname.
+    ///
+    /// Returns a channel `Receiver` to receive events about the hostname.
+    /// The caller can call `.recv_async().await` on this receiver to handle events in an
+    /// async environment or call `.recv()` in a sync environment.
+    pub fn resolve(&self, hostname: &str) -> Result<Receiver<ResolutionEvent>> {
```
keepsimple1, line 243: IMO `resolve` is generic, i.e. more than querying for addresses. Either we pass in an enum of different things to resolve, or define different methods for each kind of resolve. (Same point as I suggested to add a QueryType)

keepsimple1 (commented):
Thank you for your PR!  With a quick look, it didn't feel quite right for me. I left one inline comment, but I need more time to think through and review other parts. 

(I'm hoping to simplify the current solution and still meet your needs)

src/service_daemon.rs
```diff
@@ -2223,7 +2446,24 @@ fn check_service_name(fullname: &str) -> Result<()> {
...
 }
 
-fn call_listener(
+/// Validate a hostname.
+fn check_hostname(hostname: &str) -> Result<()> {
+    if !hostname.ends_with(".local.") {
+        return Err(e_fmt!("Hostname must end with '.local.'"));
+    }
+
+    if hostname.is_empty() {
```
keepsimple1, line 2455: This check is not needed because the previous check already made sure it's not empty

src/service_daemon.rs
```diff
@@ -1914,6 +2073,20 @@ pub enum ServiceEvent {
...
+pub enum ResolutionEvent {
+    /// Started searching for the ip address of a hostname.
+    SearchStarted(String),
+    /// One or more addresses for a hostname has been found.
+    HostnameAddressesFound(String, HashSet<IpAddr>),
+    /// One or more addresses for a hostname has been removed.
+    HostnameAddressesRemoved(String, HashSet<IpAddr>),
+    /// Stopped searching for the ip address of a hostname.
+    SearchStopped(String),
+}
```
keepsimple1, line 2088: I wonder if `String` (i.e. the hostname) is needed in each variant. If `ResolutionEvent` is only sent to a given channel created by a `resolve_*` call, then the `String` is redundant as the caller has it.  And the variant `HostnameAddressesFound` can be named as `AddressesFound`.

If we keep the `String` value, then a question is:  what happens if `ResolutionEvent` is for things other than `hostname`?   I think it's more clearly to name `SearchStarted` as `HostnameSearchStarted`.

keepsimple1 (commented):
I went through the whole diff 😅 . It looks good to me overall. Please see inline for some comments. The main thing is still regarding the public API & enum.  (That's always the hard part for a library ;-) ). 

Also, could you please add test cases for this new feature? Thanks!

src/service_daemon.rs
```diff
@@ -569,6 +625,37 @@ impl ServiceDaemon {
...
+                    error!(
+                        "Failed to send ResolveStarted({})(repeating:{}): {}",
+                        &hostname, repeating, e
+                    );
+                    return;
+                }
+                if !repeating {
+                    zc.add_hostname_querier(hostname.clone(), listener.clone());
+                    // if we already have the records in our cache, just send them
+                    zc.query_cache_for_hostname(&hostname, listener.clone());
```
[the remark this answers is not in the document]
oysteintveit-nordicsemi: I think using the cache is good. But I don't think there's going to be a lot of people running overlapping service discovery and hostname resolution threads at the same time (in the sense that they'd use the same `A`/`AAAA` records), considering the service discovery part already resolves the host for you automatically. So the value is probably miniscule.

But in the case that the host goes down inbetween the time you resolved the hostname and the time you were going to connect to the service, you'd have to handle the error while establishing a connection anyway, no matter whether you're using a 60 second old record from the cache or a record you received 20 milliiseconds beforehand. Whether a record is coming from the cache or has just been received shouldn't really matter to the user, if they're doing their error handling properly.

> If I were to resolve a hostname, I might want to always query it for the current resolution even they are already in the cache

I think it would be better to let the library user have access to a function that lets them flush either the whole, or parts of the cache. That would give them a choice to opt out, and would give them the power to do things like flushing the cache when you're having connection problems and seeing whether the host still can be resolved, to provide information about whether it's a problem with the service or the host, etc.

> How about your use case?

I'm using this for a cli tool that exits after each invocation, so the entire cache is dropped. It doesn't really affect me.

src/service_daemon.rs
oysteintveit-nordicsemi: Right, I'll check that there's something before `.local.` instead.

On a side note, should I run `check_hostname` on the hostname given to `ServiceInfo`? I had some trouble earlier while testing this, where I had given `mymachine.local` (without the last dot) to the service announcer, and then trying to resolve the hostname `mymachine.local.`

src/service_daemon.rs
oysteintveit-nordicsemi: > I wonder if `String` (i.e. the hostname) is needed in each variant. If `ResolutionEvent` is only sent to a given channel created by a `resolve_*` call, then the `String` is redundant as the caller has it.

That's right, I hadn't given it much thought. We could certainly do without it. But I see you're returning the `service_type` in all the `ServiceEvent`s as well, not sure if we should keep the same behaviour to make it consistent?

I could also see someone using it to convert several `Receive<Event>` structs into streams, and combine them into one large stream that combines the handling for all channels into a single loop, without having to do reverse lookups for figuring out which hostname the event belongs to (I'm not sure if that even works).

> And the variant `HostnameAddressesFound` can be named as `AddressesFound`.
>
> If we keep the `String` value, then a question is: what happens if `ResolutionEvent` is for things other than hostname? I think it's more clearly to name `SearchStarted` as `HostnameSearchStarted`.

I guess it depends on what you'd want to extend it to do in the future? It might be better to just name the whole enum `HostnameResolutionEvent`, and rather make other `-Event` structs for other kinds of functionality. Referring back to the decision not to merge this hostname resolution functionality with `ServiceEvent`, in order to avoid figuring out which enum variant belonged to what querier, and to avoid having the user having to handle enum variants that will never appear for the type of query they're using.

I think the only argument for keeping the hostname in the `SearchStarted` and `SearchStopped` events would be consistency with `ServiceEvent`.

What do you think?

oysteintveit-nordicsemi:
Okay, so I've added a `timeout: Option<u64>` to `ServiceDaemon:resolve_hostname`. This adds the potential timeout to a table of timeouts, `Zeroconf::hostname_resolver_timeouts: HashMap<String, u64> // (hostname, timeout)`, marks it in timers, and the `Command::ResolveHostname` continues to add retransmissions until it has reached the timeout. A block in the runloop will then clean up the tables, and notify the listener of both the timeout and the search stop.

src/service_daemon.rs
```diff
@@ -910,6 +1077,11 @@ struct Zeroconf {
     /// Socket for signaling.
     signal_sock: UdpSocket,
 
+    /// Timestamps marking where we need another iteration of the run loop,
+    /// to react to events like retransmissions, cache refreshes, interface IP address changes, etc.
+    ///
+    /// When the run loop goes through a single iteration, it will
+    /// set its timeout to the earliest timer in this list.
     timers: Vec<u64>,
```
oysteintveit-nordicsemi, line 1085: This part took me some time to understand, so I've added a comment. I'm still not 100% sure that I understood it properly, so I want to point it out for discussion. Please suggest a better description or corrections if you have any

src/service_daemon.rs
keepsimple1: I think it's good to run the same check.  It might triggers a few test case failures though as we didn't enforce it so far. If you have time to make the changes, that will be great! Otherwise, it's perfectly fine if you open a new issue to track it.

src/service_daemon.rs
keepsimple1: > But I see you're returning the `service_type` in all the `ServiceEvent`s as well, not sure if we should keep the same behaviour to make it consistent?

Good point! I agree it's better to keep the the hostname.

> I guess it depends on what you'd want to extend it to do in the future? It might be better to just name the whole enum HostnameResolutionEvent, and rather make other -Event structs for other kinds of functionality.

I see that `avahi` supports `avahi-resolve-address` , i.e. reverse lookup hostname. I don't think if we will support it in future. And again, I further don't know if it's good idea to extend the same `ResolutionEvent` for that case.

All things considered, I think your idea of just naming it `HostnameResolutionEvent` is good.  And add `#[non_exhaustive]` just in case.

src/service_daemon.rs
```diff
@@ -888,7 +1046,16 @@ struct Zeroconf {
...
+    service_queriers: HashMap<String, Sender<ServiceEvent>>, // <ty_domain, channel::sender>
+
+    /// Active "ResolveHostname" commands.
+    hostname_resolvers: HashMap<String, Sender<ResolutionEvent>>, // <hostname, channel::sender>
+
+    /// Timeout timestamps for active "ResolveHostname" commands.
+    ///
+    /// The timestamps are set at the future time when the command should timeout.
+    /// If a hostname resolver does not have a timeout, it will not be added to this map.
+    hostname_resolver_timeouts: HashMap<String, u64>, // <hostname, UNIX timestamp in millis>
```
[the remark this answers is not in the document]
keepsimple1: I think it's unifying them is better. One advantage is that no chance to have inconsistent keys between two HashMaps.

keepsimple1 (commented):
Thanks for the updates!

keepsimple1:
Btw, I added support for [sending multiple questions in one query](https://github.com/keepsimple1/mdns-sd/pull/194). You can try out the new `send_query_vec` .

src/service_daemon.rs
```diff
@@ -569,6 +659,47 @@ impl ServiceDaemon {
...
+                // Only add retransmission if it does not exceed the hostname resolver timeout, if any.
+                if zc
+                    .hostname_resolver_timeouts
+                    .get(&hostname)
+                    .map(|timeout| next_time < *timeout)
+                    .unwrap_or(true)
+                {
+                    zc.add_retransmission(
+                        next_time,
+                        Command::ResolveHostname(hostname, delay, listener, timeout),
```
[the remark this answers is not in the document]
oysteintveit-nordicsemi: I think this already does what you're describing. `add_hostname_resolver` is only called on the first instance of the command due to the repeating flag, so the timeout will be from the moment the caller issue the command.

But it's not very transparent to the reader that all of this happens in `add_hostname_resolver`. Should I move it to another function (`add_hostname_resolver_timeout` or something), and call it below `add_hostname_resolver` in the non-repeat block? I can also set `timeout` to `None` in the retransmissions since the data is no longer needed, and to avoid further confusion.

src/service_daemon.rs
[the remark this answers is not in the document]
oysteintveit-nordicsemi: I've unified them using `HashMap<String, (Sender<ResolutionEvent>, Option<u64>)>` (note the `Option`)

src/service_daemon.rs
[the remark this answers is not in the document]
oysteintveit-nordicsemi: I've set the timeout to `None` in the retransmissions, but kept the rest in `add_hostname_resolver`. Let me know if you would like me to change it

src/service_daemon.rs
oysteintveit-nordicsemi: I have added the hostname check to registered services, and fixed the tests. But note that it might be a breaking change for some users.

src/service_daemon.rs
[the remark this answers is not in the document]
keepsimple1: 👍  this is good

src/service_daemon.rs
keepsimple1: Thanks for doing this extra! You're right, this will be a breaking change. I will document this in the next version release. It's a trade-off but I think overall it's good to change it. (If any surprise happens, we will handle it ;-))

src/service_daemon.rs
```diff
@@ -1920,6 +2148,23 @@ pub enum ServiceEvent {
...
+/// regarding host resolution.
+#[derive(Debug)]
+#[non_exhaustive]
+pub enum HostnameResolutionEvent {
+    /// Started searching for the ip address of a hostname.
+    SearchStarted(String),
+    /// One or more addresses for a hostname has been found.
+    HostnameAddressesFound(String, HashSet<IpAddr>),
+    /// One or more addresses for a hostname has been removed.
+    HostnameAddressesRemoved(String, HashSet<IpAddr>),
```
keepsimple1, line 2161: I felt that `Hostname` prefix could be removed from `HostnameAddressesFound` and `HostnameAddressesRemoved` as the enum name already included `Hostname`. 

That said, I am not super good at naming. So it's only a personal opinion.

src/service_daemon.rs
```diff
@@ -954,7 +1121,9 @@ impl Zeroconf {
             poll_id_count: 0,
             my_services: HashMap::new(),
             cache: DnsCache::new(),
-            queriers: HashMap::new(),
+            hostname_resolvers: HashMap::new(),
+            // hostname_resolver_timeouts: HashMap::new(),
```
keepsimple1, line 1125: nitpick:  this leftover shall be removed.

keepsimple1 (commented):
Thank you for updating! Looks good! I only have one minor inline comment regarding naming.

keepsimple1 (approved):
LGTM! Thanks!

PR #196, opened 2024-04-15
zeroconf: use min heap for timers

oysteintveit-nordicsemi:
Small improvement, use `Collections::BinaryHeap` to keep track of the order of the timers.

I've built this change on top of #192 because it will conflict, undrafting once that is completed.


keepsimple1 (approved):
👍 pretty cool!  LGTM!

PR #198, opened 2024-04-17
Fix flaky test

oysteintveit-nordicsemi:
Seems like the `subtype` test has become flaky, after I accidentally named the service equally for the hostname resolution test from #192. This PR fixes that, as well as adding some dependencies for capturing output logs made by the `log` macros.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,6 +24,7 @@ polling = "2.1"                                        # select/poll sockets
 socket2 = { version = "0.5.5", features = ["all"] }      # socket APIs
 
 [dev-dependencies]
-env_logger = "0.11.3"
+env_logger = { version = "= 0.10.2", default-features = false, features= ["humantime"] }
 fastrand = "1.8"
-test-log = "0.2.15"
+test-log = "= 0.2.14"
+test-log-macros = "= 0.2.14"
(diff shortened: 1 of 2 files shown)


oysteintveit-nordicsemi:
Seems like the MSRV of `env_logger` version 0.11.3 is 1.71. Should I remove the extra test dependencies in favor of keeping this project's MSRV to 1.63.0?

Cargo.toml
```diff
@@ -24,4 +24,6 @@ polling = "2.1"                                        # select/poll sockets
 socket2 = { version = "0.5.5", features = ["all"] }      # socket APIs
 
 [dev-dependencies]
+env_logger = "0.11.3"
 fastrand = "1.8"
+test-log = "0.2.15"
```
keepsimple1, line 29: First time seeing this crate. Seems neat based on its doc. Is it working well for you?

keepsimple1 (commented):
Thanks for fixing the test! Minor comments inline.

Cargo.toml
oysteintveit-nordicsemi: This one was tricky to downgrade for some reason. Cargo kept insisting that we use version 0.2.15 of transitive dependency `test-log-macros`, despite `test-log 0.2.14` depending on version `test-log-macros 0.2.14` itself. I've included `test-log-macros` as a dev dependency just to trick cargo to use the right one, but we never use that dependency directly.

I have not used this crate before myself, but I've found myself adding `env_logger::init()` to the start of every single test I write. I'll probably start using it from now on.

keepsimple1 (approved):
LGTM! Thanks!

PR #201, opened 2024-04-24
feat: honor cache flush

lyager:
According to RFC 6762, when the cache-flush bit is set on a RR, all previously cache entries should be set to expire 1 second in the future.

This bit was previously called 'UNIQUE` which might have the same meaning, but cache-flush seems more in line with the RFC

### DIFF
--- a/src/service_daemon.rs
+++ b/src/service_daemon.rs
@@ -2363,7 +2363,7 @@ impl DnsCache {
                 // When cache flush is asked, we set expire date to 1 second in the future
                 // if created more than 1 second ago
                 // Ref: RFC 6762 Section 10.2
-                if now > r.get_created() + 1000 {
+                if incoming.get_class() == r.get_class() && now > r.get_created() + 1000 {
                     r.set_expire(now + 1000);
                 }
             });
@@ -2673,7 +2673,6 @@ mod tests {
             FLAGS_QR_RESPONSE, TYPE_PTR, TYPE_TXT,
         },
         service_daemon::check_hostname,
-        service_info::IntoTxtProperties,
     };
     use std::{collections::HashMap, net::SocketAddr, net::SocketAddrV4, time::Duration};
 
--- a/src/service_daemon.rs
+++ b/src/service_daemon.rs
@@ -2363,7 +2363,7 @@ impl DnsCache {
                 // When cache flush is asked, we set expire date to 1 second in the future
                 // if created more than 1 second ago
                 // Ref: RFC 6762 Section 10.2
-                if now - r.get_created() > 1000 {
+                if now > r.get_created() + 1000 {
                     r.set_expire(now + 1000);
                 }
             });
--- a/src/service_daemon.rs
+++ b/src/service_daemon.rs
@@ -2673,6 +2673,7 @@ mod tests {
             FLAGS_QR_RESPONSE, TYPE_PTR, TYPE_TXT,
         },
         service_daemon::check_hostname,
+        service_info::IntoTxtProperties,
     };
     use std::{collections::HashMap, net::SocketAddr, net::SocketAddrV4, time::Duration};
 
--- a/src/service_daemon.rs
+++ b/src/service_daemon.rs
@@ -33,7 +33,7 @@ use crate::log::{debug, error, warn};
 use crate::{
     dns_parser::{
         current_time_millis, DnsAddress, DnsIncoming, DnsOutgoing, DnsPointer, DnsRecordBox,
-        DnsRecordExt, DnsSrv, DnsTxt, CLASS_IN, CLASS_UNIQUE, FLAGS_AA, FLAGS_QR_QUERY,
+        DnsRecordExt, DnsSrv, DnsTxt, CLASS_CACHE_FLUSH, CLASS_IN, FLAGS_AA, FLAGS_QR_QUERY,
         FLAGS_QR_RESPONSE, MAX_MSG_ABSOLUTE, TYPE_A, TYPE_AAAA, TYPE_ANY, TYPE_NSEC, TYPE_PTR,
         TYPE_SRV, TYPE_TXT,
     },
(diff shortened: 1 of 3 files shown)


lyager:
I've had issues with embedded devices changing IP's and appearing with both the previous and current IP. I'm guessing this goes for other records too?

src/service_daemon.rs
```diff
@@ -2356,6 +2356,15 @@ impl DnsCache {
             _ => return None,
         };
 
+        if incoming.get_cache_flush() {
+            // Mark all existing records of this type as expired, and prepend the new record.
+            record_vec.iter_mut().for_each(|r| {
+                r.expire();
```
keepsimple1, line 2362: Per RFC 6762,  "...all old records with that name, rrtype,
   and rrclass that were received more than one second ago are declared
   invalid, and marked to expire from the cache in one second."

we should check for the same rrclass here,  and check for "received more than one second ago"  (You can use `DnsRecord.created` timestamp to check) .

keepsimple1 (commented):
Thanks for your PR!  Please see comments inline. And could you please add tests for the new feature?

lyager:
Thanks for the response, I'll dive into it 👍

src/service_daemon.rs
lyager: Checking when it was created, added `get_created()`

lyager:
I had some trouble constructing a proper test. My idea was to start browsing, similar to the test `service_with_temporarily_invalidated_ptr`, but I had trouble constructing response packets. I'll give it another shot, but if you have some test code where MDNS response packets are constructed and send out, please let me know :)

src/service_daemon.rs
```diff
@@ -2356,6 +2356,21 @@ impl DnsCache {
             _ => return None,
         };
 
+        if incoming.get_cache_flush() {
+            // Mark all existing records of this type as expired, and prepend the new record.
+            record_vec.iter_mut().for_each(|r| {
+                // When cache flush is asked, we set expire date to 1 second in the future
+                // if created more than 1 second ago
+                // Ref: RFC 6762 Section 10.2
+                let now = current_time_millis();
```
keepsimple1, line 2365: I think it's better to only call `current_time_millis` once before entering `for_each` loop. The time accuracy loss should near 0 (for millisecond)

keepsimple1:
> I had some trouble constructing a proper test. My idea was to start browsing, similar to the test `service_with_temporarily_invalidated_ptr`, but I had trouble constructing response packets. I'll give it another shot, but if you have some test code where MDNS response packets are constructed and send out, please let me know :)

`service_with_temporarily_invalidated_ptr` is a good sample to start. The response packet is constructed in `to_packet_data` called by `broadcast_dns_on_intf` .  And you want to add `CLASS_CACHE_FLUSH` when creating a record (valid record instead of the invalid one in below): 

https://github.com/keepsimple1/mdns-sd/blob/1b9647d8c90986e2ee14d31d6907e38b516c26be/src/service_daemon.rs#L2775-L2784

src/service_daemon.rs
lyager: You're right. Moved it.

lyager:
I fixed the already existing test. Thanks for the help with understanding, I'll attempt writing a specific cache-flush test now.

lyager:
Test added for cache flush - hope it's ok.

src/service_daemon.rs
```diff
@@ -2816,4 +2835,135 @@ mod tests {
...
+
+        resolved = false;
+        while let Ok(event) = browse_chan.recv_timeout(timeout) {
+            match event {
+                ServiceEvent::ServiceResolved(info) => {
+                    resolved = true;
+                    println!("Resolved a service of {}", &info.get_fullname());
+                    // Ensure that only the new property remains after cache flush, and it's the
+                    // newly defined txt record from `new_service`
+                    // assert_eq!(new_properties.into_txt_properties(), *info.get_properties());
```
keepsimple1, line 2957: Shouldn't this check be uncommented?

keepsimple1 (commented):
Thanks for your updates. Some minor comments inline.

src/service_daemon.rs
```diff
@@ -2356,6 +2356,21 @@ impl DnsCache {
...
         };
 
+        if incoming.get_cache_flush() {
+            // Mark all existing records of this type as expired, and prepend the new record.
+            let now = current_time_millis();
+            record_vec.iter_mut().for_each(|r| {
+                // When cache flush is asked, we set expire date to 1 second in the future
+                // if created more than 1 second ago
+                // Ref: RFC 6762 Section 10.2
+                if now - r.get_created() > 1000 {
```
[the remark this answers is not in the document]
lyager: I guess when Unix timestamp runs over we will be in bigger trouble - but yeah. Again, thanks for the feedback :)

tests/mdns_test.rs
```diff
@@ -146,13 +146,13 @@ fn integration_success() {
     let count = addr_count.lock().unwrap();
     assert_eq!(*count, my_addrs_count);
 
-    // `resolve_count` is not guaranteed to always be 1
-    // or `my_addrs_count`. If `my_addrs_count` > 1, these
-    // addrs could be resolved in a single message from
-    // the daemon, or in separate messages.
-    let count = resolve_count.lock().unwrap();
-    assert!(*count >= 1);
-    assert!(*count <= my_addrs_count);
```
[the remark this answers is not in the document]
lyager: Exactly.. Since the cache_flush bit will trigger a new ServiceResolved - and cache_flush bit is being set by the client, the number in `resolve_count` would be higher now. So I introduced the HashSet.

src/service_daemon.rs
```diff
@@ -2356,6 +2356,21 @@ impl DnsCache {
...
+                    r.set_expire(now + 1000);
+                }
+            });
+            record_vec.insert(0, incoming);
```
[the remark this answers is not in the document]
lyager: I was wondering the same. Re-reading the RFC it seems like the right way to do this, since we are looking at the records one at a time, and the assumption in the RFC looks to be everything below a second can be part of a burst.

lyager:
Thanks for the feedback, I've stoppede squashing the commit, if you prefer me to, let me know.

keepsimple1 (approved):
LGTM, thanks !

PR #205, opened 2024-05-01
feat: extract match addr to type as a function

CosminPerRam:
I have seen the repeated usage of
```rust
match address {
    IpAddr::V4(_) => TYPE_A,
    IpAddr::V6(_) => TYPE_AAAA,
}
```
in multiple places, although this is a very simple operation, it could be extracted into a function (which is what this PR does).
I'm not really sure if the function name `ip_address_to_type` and its location are appropriate.

### DIFF
--- a/src/dns_parser.rs
+++ b/src/dns_parser.rs
@@ -52,6 +52,7 @@ pub const FLAGS_AA: u16 = 0x0400; // mask for Authoritative answer bit
 
 pub type DnsRecordBox = Box<dyn DnsRecordExt + Send>;
 
+#[inline]
 pub const fn ip_address_to_type(address: IpAddr) -> u16 {
     match address {
         IpAddr::V4(_) => TYPE_A,
--- a/src/dns_parser.rs
+++ b/src/dns_parser.rs
@@ -53,7 +53,7 @@ pub const FLAGS_AA: u16 = 0x0400; // mask for Authoritative answer bit
 pub type DnsRecordBox = Box<dyn DnsRecordExt + Send>;
 
 #[inline]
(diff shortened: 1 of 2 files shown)


src/dns_parser.rs
```diff
@@ -52,6 +52,14 @@ pub const FLAGS_AA: u16 = 0x0400; // mask for Authoritative answer bit
 
 pub type DnsRecordBox = Box<dyn DnsRecordExt + Send>;
 
+#[inline]
+pub const fn ip_address_to_type(address: IpAddr) -> u16 {
```
keepsimple1, line 56: nitpick:  do you mind pass `address` as `&IpAddr` ?

src/service_daemon.rs
```diff
@@ -565,10 +565,7 @@ impl ServiceDaemon {
                 for (hostname, ip_addr) in
                     zc.cache.refresh_due_hostname_resolutions(hostname).iter()
                 {
-                    match ip_addr {
-                        IpAddr::V4(_) => zc.send_query(hostname, TYPE_A),
-                        IpAddr::V6(_) => zc.send_query(hostname, TYPE_AAAA),
-                    }
+                    zc.send_query(hostname, ip_address_to_type(*ip_addr));
```
keepsimple1, line 568: If we pass `&IpAddr` for the function, then no need to have `*` here.

keepsimple1 (commented):
Thanks for your PR! It's a nice refactoring! Minor comments inline.

keepsimple1 (approved):
Nice! LGTM, thanks!

PR #206, opened 2024-05-01
feat: replace remaining box dns with declared type

CosminPerRam:
After skimming through the code a bit, I found two usages of `Box<dyn DnsRecordExt>` instead of using the `DnsRecordBox` type (`Box<dyn DnsRecordExt + Send>`).
Although in the replaced remaining cases, we dont really use the `Send` bound but this can be done for consistency sake as there are already other usages of it in places where it isn't necessarily needed.


keepsimple1 (approved):
LGTM!

PR #207, opened 2024-05-01
perf: in adding answers, use static dispatch instead of dynamic dispatch

CosminPerRam:
This is an alternative to #206, changes the `answer` type from `Box<dyn DnsRecordExt>` to `impl DnsRecordExt + Send + 'static`.

This replaces the usage of dynamic dispatch on every function call with static dispatch, which in our case should be faster as we always call these functions with new instances (`Box::new(...)`), so the compiler should be able to monomorphize and/or optimize further its usage, as all the possible implementations are now known at compile time.

The dynamic dispatch is still needed for the `answer` field (`Vec<(Box<dyn DnsRecordExt>, u64)>`) in `self.answers.push` of `add_answer_at_time`, but this is done only if needed (when if conditions are met), so this should provide us with only performance improvements, please note that I'm not really a Rust expert or anything, nor have I done benchmarks to provide this (maybe we could add some eventually?).

Edit: forgot to mention that the disadvantages of static dispatch are longer compile times and larger binary size (but I haven't seen significant changes in these)

### DIFF
--- a/src/dns_parser.rs
+++ b/src/dns_parser.rs
@@ -50,7 +50,7 @@ pub const FLAGS_QR_QUERY: u16 = 0x0000;
 pub const FLAGS_QR_RESPONSE: u16 = 0x8000;
 pub const FLAGS_AA: u16 = 0x0400; // mask for Authoritative answer bit
 
-pub type DnsRecordBox = Box<dyn DnsRecordExt>;
+pub type DnsRecordBox = Box<dyn DnsRecordExt + Send>;
 
 #[inline]
 pub const fn ip_address_to_type(address: &IpAddr) -> u16 {
@@ -779,9 +779,9 @@ impl DnsOutgoing {
     //    server/responder SHOULD include the following additional records:
 
     //    o  All address records (type "A" and "AAAA") named in the SRV rdata.
-    pub(crate) fn add_additional_answer(&mut self, answer: DnsRecordBox) {
+    pub(crate) fn add_additional_answer(&mut self, answer: impl DnsRecordExt + Send + 'static) {
         debug!("add_additional_answer: {:?}", &answer);
-        self.additionals.push(answer);
+        self.additionals.push(Box::new(answer));
     }
 
     /// Returns true if `answer` is added to the outgoing msg.
@@ -789,7 +789,7 @@ impl DnsOutgoing {
     pub(crate) fn add_answer(
         &mut self,
         msg: &DnsIncoming,
-        answer: impl DnsRecordExt + 'static,
+        answer: impl DnsRecordExt + Send + 'static,
     ) -> bool {
         debug!("Check for add_answer");
         if !answer.suppressed_by(msg) {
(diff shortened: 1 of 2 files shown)


src/dns_parser.rs
```diff
@@ -158,7 +158,7 @@ impl PartialEq for DnsRecord {
     }
 }
 
-pub trait DnsRecordExt: fmt::Debug {
+pub trait DnsRecordExt: fmt::Debug + Send {
```
CosminPerRam, line 161: The compiler automatically implements `Send` on appropriate types, but as we have only usages of this now, it's a little bit more concise to explicitly state it on the trait rather than on the Box (and not in the answer functions), theoretically at least, should be the same thing but I think stating it like this rather than how it was is a bit better, what do you think?

keepsimple1:
On the grand scheme, I think the performance difference is small / not significant for this lib, unless we have some measurement data showing a problem.  But It's a learning opportunity for me to understand more about dynamic dispatch and static dispatch in Rust. 

> This replaces the usage of dynamic dispatch on every function call with static dispatch, 

For my understanding, do you have a pointer to Rust doc or something similar to verify that? I don't know for sure, and asked ChatGPT, and here is what it says: 

`...even with the impl Trait syntax, dynamic dispatch will still be used when calling the add_additional_answer function because answer is still treated as a trait object. ...`

src/dns_parser.rs
keepsimple1: I think it is odd to have `DnsRecordExt` trait to conform `Send`, because this trait is really just about supporting different types of DNS records, nothing related to dispatching mechanisms (e.g. Send or Sync, etc).

I can understand that moving `Send` to the trait would simplify the parameter signature in functions, but it doesn't feel right for me.

src/dns_parser.rs
```diff
@@ -771,14 +771,18 @@ impl DnsOutgoing {
     //    server/responder SHOULD include the following additional records:
 
     //    o  All address records (type "A" and "AAAA") named in the SRV rdata.
-    pub(crate) fn add_additional_answer(&mut self, answer: DnsRecordBox) {
+    pub(crate) fn add_additional_answer(&mut self, answer: impl DnsRecordExt + 'static) {
```
keepsimple1, line 774: Although I'm not sure about the `static dispatch` thing, I see the benefit of not requiring the callers to add `Box::new()` at the call site.

And I think I prefer to have `Send` in here, compared to move `Send` into `DnsRecordExt` trait, as I mentioned in another comment.

CosminPerRam:
> Is ChatGPT hallucinating here?

Asking ChatGPT (4) about this [says that this statement is incorrect](https://chat.openai.com/share/7ab00fdf-605b-465e-82c2-4e7def035078), I think it says that it still implies dynamic dispatch as we eventually use `.push(Box::new(answer))`.

[Here](https://www.cs.brandeis.edu/~cs146a/rust/doc-02-21-2015/book/static-and-dynamic-dispatch.html) is how Brandeis University explains what is static and dynamic dispatch and their advantages and disadvantages. [Here](https://reintech.io/blog/understanding-implementing-rust-static-dynamic-dispatch) is a shorter example and explanation. 

> I see the benefit of not requiring the callers to add Box::new() at the call site. So I'm okay with such change, for ease of use, not for the performance reason.

Fair enough, if I will get the time I'll try to make some benchmarks just as a curiosity (:

keepsimple1 (approved):
LGTM, thanks!

PR #208, opened 2024-05-01
feat: remove clone derive from counter

CosminPerRam:
Just a very small nitpick I found (:
For the `Counter` enum, the derive `Clone` isn't used at all, it was added in [this commit](https://github.com/CosminPerRam/mdns-sd/commit/c5a759e9667a904db65b0c8fe9375ed3e02ab26d) for reference.


keepsimple1 (approved):
LGTM!

PR #209, opened 2024-05-03
feat: merge match same arms

CosminPerRam:
Title, there were 2 instances of match arms being the same (A and AAAA).


keepsimple1 (approved):
Nice!

PR #210, opened 2024-05-08
Bugfix: AddressesRemoved event does not contain removed addrs

keepsimple1:
Currently when `HostnameResolutionEvent::AddressesRemoved` event was sent, the addresses included are the ones still in use. They should be the addressses got removed. 

A test case is added to verify the addresses.

@oysteintveit-nordicsemi if you have time, please check if I missed anything.

~~(The cache flush test seems flaky, will fix as well)~~


keepsimple1:
Rebased to the latest main branch.

Issue #193, opened 2024-04-12
Add support for traffic reduction techniques from RFC6762, sec. 7


oysteintveit-nordicsemi:
(I made a small comment about this in #191, but I thought it would be better to create a proper issue)

RFC6762 includes a whole section on techniques to reduce network traffic. The current implementation works fine without these techniques - the daemon will properly answer anyone requesting service details, and it's able to query for service details just fine. However, it is wasteful and it's technically not compliant with the RFC due to the MUST requirements.

I'll include some extracts of the contents in the RFC, and roughly what needs to be done in the points below.

### [Sec. 7.1 - Known Answer Suppression](https://datatracker.ietf.org/doc/html/rfc6762#section-7.1) (MUST)

> When a Multicast DNS querier sends a query to which it already knows
   some answers, it populates the Answer Section of the DNS query
   message with those answers.

I see there's a method [`DnsOutgoing::add_additional_answer`](https://github.com/keepsimple1/mdns-sd/blob/c141cb370f373e21de0bb672753996beba85de4b/src/dns_parser.rs#L751) that seems to be able to write answers to any question packet, so I believe the only missing piece is to fetch any relevant non-expired records from the dns cache, and add it to any outgoing packets.

### [Sec. 7.2 - Multipacket Known Answer Suppression](https://datatracker.ietf.org/doc/html/rfc6762#section-7.2) (MUST)

> Sometimes a Multicast DNS querier will already have too many answers
   to fit in the Known-Answer Section of its query packets.  In this
   case, it should issue a Multicast DNS query containing a question and
   as many Known-Answer records as will fit.  It MUST then set the TC
   (Truncated) bit in the header before sending the query.  It MUST
   immediately follow the packet with another query packet containing no
   questions and as many more Known-Answer records as will fit.  [...]
>
>  A Multicast DNS responder seeing a Multicast DNS query with the TC
   bit set defers its response for a time period randomly selected in
   the interval 400-500 ms.  [...]
 
[...]

PR #232, opened 2024-07-08
add support for Known Answer Suppression part 2: multi-packet

keepsimple1:
This is to address #193 section 2: multiple packets

### DIFF
--- a/src/dns_parser.rs
+++ b/src/dns_parser.rs
@@ -56,7 +56,7 @@ pub const FLAGS_AA: u16 = 0x0400;
 
 pub const FLAGS_TC: u16 = 0x0200; // TC(Truncated) bit
 
-pub type DnsRecordBox = Box<dyn DnsRecordExt>;
+pub(crate) type DnsRecordBox = Box<dyn DnsRecordExt>;
 
 #[inline]
 pub const fn ip_address_to_type(address: &IpAddr) -> u16 {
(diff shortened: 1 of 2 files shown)


keepsimple1:
There is a pending issue with the changes on the Responder side:

Per [RFC 6762 section 7.2](https://datatracker.ietf.org/doc/html/rfc6762#section-7.2): 
```
A Multicast DNS responder seeing a Multicast DNS query with the TC
   bit set defers its response for a time period randomly selected in
   the interval 400-500 ms.  This gives the Multicast DNS querier time
   to send additional Known-Answer packets before the responder

   responds.  If the responder sees any of its answers listed in the
   Known-Answer lists of subsequent packets from the querying host, it
   MUST delete that answer from the list of answers it is planning to
   give (provided that no other host on the network has also issued a
   query for that record and is waiting to receive an answer).
```

The above means the responder needs to know the sender of each packet (e.g. the queryer IP address), otherwise it cannot associate the additional Known-Answer lists of the subsequent packets with a sender (and the question).
 
However `socket2` lib current version (0.5.5) does not have a safe API to get the source address of a packet, and I don't want to use any unsafe APIs.

There are a couple of options I'm thinking of:  1) not to implement the Responder side for multi-packet.  2) figure out a basic / narrower case for the Responder that handles certain cases of multi-packet without using the source address.

keepsimple1:
Because of the issue describe in earlier comments, I decided to merge in the support for the querier side first. This will unblock Known-Answer feature, except not supporting the multi-packet on the responder side. I think it is better than keep all of them stuck in PR.

PR #239, opened 2024-08-12
Avoid redundant query, announcement and unregistration overhaul

hrzlgnm:
I had a second thought about the "already sent query or announcemnet in the subnet approach".

Using a "subnet" to filter already notified multicast groups on IPv4 or IPv6 is not a effective method in a multicast context.

The socket used to send multicast traffic is acutally bound to the "any" address associated with the respective address family and is joined on the mDNS multicast_v(4|6) address using the network interface index. 

If we have multiple IPv6 or IPv4 addresses on the same network interface in different **subnets**, this will result in redundant traffic to be sent when sending queries, unregistration messages or announcements, as we still sent those to the same multicast group on the same interface.

I've came up with another approach. We use a tuple(interface index , adress family ) in a set to track already "used" notifications.

This reduces the number of messages sent drastically for example on my windows machine, i'm getting 3 IPv6 addresses and one IPv4 address. Before this change this would result in 4 messages being send on that interface. With this change we only send 2.

@keepsimple1 Please, let me know if there are any problems with the approach.

### DIFF
--- a/src/service_daemon.rs
+++ b/src/service_daemon.rs
@@ -1393,7 +1393,7 @@ impl Zeroconf {
 
     /// Sends a multicast query for `name` with `qtype`.
     fn send_query(&self, name: &str, qtype: u16) {
-        self.send_query_vec(&[(name, qtype)])
+        self.send_query_vec(&[(name, qtype)]);
     }
 
     /// Sends out a list of `questions` (i.e. DNS questions) via multicast.
--- a/src/service_daemon.rs
+++ b/src/service_daemon.rs
@@ -39,11 +39,11 @@ use crate::{
         TYPE_PTR, TYPE_SRV, TYPE_TXT,
     },
     error::{Error, Result},
-    service_info::{ifaddr_subnet, ServiceInfo},
+    service_info::{ServiceInfo},
     Receiver,
 };
 use flume::{bounded, Sender, TrySendError};
-use if_addrs::Interface;
+use if_addrs::{IfAddr, Interface};
 use polling::Poller;
 use socket2::{SockAddr, Socket};
 use std::{
@@ -111,6 +111,8 @@ enum Counter {
     CacheRefreshSRV,
     CacheRefreshAddr,
     KnownAnswerSuppression,
+    QuerySuppression,
+    AnnouncementSuppression,
 }
 
 impl fmt::Display for Counter {
@@ -127,6 +129,8 @@ impl fmt::Display for Counter {
             Self::CacheRefreshSRV => write!(f, "cache-refresh-srv"),
             Self::CacheRefreshAddr => write!(f, "cache-refresh-addr"),
             Self::KnownAnswerSuppression => write!(f, "known-answer-suppression"),
+            Self::QuerySuppression => write!(f, "query-suppression"),
+            Self::AnnouncementSuppression => write!(f, "announcement-suppression"),
         }
     }
 }
@@ -570,16 +574,19 @@ impl ServiceDaemon {
 
             // Refresh cached A/AAAA records with active queriers
             let mut query_count = 0;
+            let mut supress_count = 0i64;
             for (hostname, _sender) in zc.hostname_resolvers.iter() {
                 for (hostname, ip_addr) in
                     zc.cache.refresh_due_hostname_resolutions(hostname).iter()
                 {
-                    zc.send_query(hostname, ip_address_to_type(ip_addr));
+                    let sent = zc.send_query(hostname, ip_address_to_type(ip_addr));
                     query_count += 1;
+                    supress_count += zc.intf_socks.len() as i64 - sent;
                 }
             }
 
             zc.increase_counter(Counter::CacheRefreshAddr, query_count);
+            zc.increase_counter(Counter::QuerySuppression, supress_count);
 
             // check and evict expired records in our cache
             let now = current_time_millis();
@@ -1180,6 +1187,8 @@ impl Zeroconf {
         }
 
         let outgoing_addrs = self.send_unsolicited_response(&info);
+        let suppressed = (self.intf_socks.len() - outgoing_addrs.len()) as i64;
+        self.increase_counter(Counter::AnnouncementSuppression, suppressed);
         if !outgoing_addrs.is_empty() {
             self.notify_monitors(DaemonEvent::Announce(
                 info.get_fullname().to_string(),
@@ -1199,21 +1208,25 @@ impl Zeroconf {
         self.my_services.insert(service_fullname, info);
     }
 
-    /// Sends out annoucement of `info` on every valid interface.
-    /// Returns the list of interface IPs that sent out the annoucement.
+    /// Sends out announcement of `info` on every valid interface.
+    /// Returns the list of interface IPs that sent out the announcement.
     fn send_unsolicited_response(&self, info: &ServiceInfo) -> Vec<IpAddr> {
         let mut outgoing_addrs = Vec::new();
-        let mut subnet_set: HashSet<u128> = HashSet::new();
+        // Send the announcement on one interface per address family.
+        let mut intf_af_set: HashSet<(u32, u8)> = HashSet::new();
 
         for (_, intf_sock) in self.intf_socks.iter() {
-            let subnet = ifaddr_subnet(&intf_sock.intf.addr);
-            if !intf_sock.intf.is_link_local() && subnet_set.contains(&subnet) {
-                continue; // no need to send again in the same subnet.
+            let af = match intf_sock.intf.addr {
+                IfAddr::V4(_) => 4u8,
+                IfAddr::V6(_) => 6u8,
+            };
+            let intf_af = (intf_sock.intf.index.unwrap_or(0), af);
+            if intf_af_set.contains(&intf_af) {
+                debug!("Already used {:#?}", intf_af);
+                continue; // no need to send again on the same interface with same address family.
             }
             if self.broadcast_service_on_intf(info, intf_sock) {
-                if !intf_sock.intf.is_link_local() {
-                    subnet_set.insert(subnet);
-                }
+                intf_af_set.insert(intf_af);
                 outgoing_addrs.push(intf_sock.intf.ip());
             }
         }
@@ -1382,12 +1395,12 @@ impl Zeroconf {
     }
 
     /// Sends a multicast query for `name` with `qtype`.
-    fn send_query(&self, name: &str, qtype: u16) {
-        self.send_query_vec(&[(name, qtype)]);
+    fn send_query(&self, name: &str, qtype: u16) -> i64 {
+        self.send_query_vec(&[(name, qtype)])
     }
 
     /// Sends out a list of `questions` (i.e. DNS questions) via multicast.
-    fn send_query_vec(&self, questions: &[(&str, u16)]) {
+    fn send_query_vec(&self, questions: &[(&str, u16)]) -> i64 {
         debug!("Sending query questions: {:?}", questions);
         let mut out = DnsOutgoing::new(FLAGS_QR_QUERY);
         let now = current_time_millis();
@@ -1403,18 +1416,25 @@ impl Zeroconf {
             }
         }
 
-        // Send the query on one interface per subnet.
-        let mut subnet_set: HashSet<u128> = HashSet::new();
+        // Send the query on one interface per address family.
+        let mut intf_af_set: HashSet<(u32, u8)> = HashSet::new();
+        let mut sent = 0i64;
         for (_, intf_sock) in self.intf_socks.iter() {
-            if !intf_sock.intf.is_link_local() {
-                let subnet = ifaddr_subnet(&intf_sock.intf.addr);
-                if subnet_set.contains(&subnet) {
-                    continue; // no need to send query the same subnet again.
-                }
-                subnet_set.insert(subnet);
+            let af = match intf_sock.intf.addr {
+                IfAddr::V4(_) => 4u8,
+                IfAddr::V6(_) => 6u8,
+            };
+            let intf_af = (intf_sock.intf.index.unwrap_or(0), af);
+            if intf_af_set.contains(&intf_af) {
+                debug!("Already used {:#?}", intf_af);
+                continue; // no need to send query the same interface with same address family.
             }
+            intf_af_set.insert(intf_af);
             send_dns_outgoing(&out, intf_sock);
+            sent += 1
         }
+
+        sent
     }
 
     /// Reads from the socket of `ip`.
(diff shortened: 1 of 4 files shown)


keepsimple1:
It's a good question, and I need to dig more as well. Here is my thoughts so far: 

> The socket used to send multicast traffic is acutally bound to the "any" address associated with the respective address family and is joined on the mDNS multicast_v(4|6) address using the network interface index.

The "bound" part is true, although for "join" it's not using the network interface index for IPv4. It's [using the IP address](https://github.com/keepsimple1/mdns-sd/blob/4772bd199a55da89b49d6e8757995641e4f20b3f/src/service_daemon.rs#L692).  IPv6 seems different.

> 
> If we have multiple IPv6 or IPv4 addresses on the same network interface in different **subnets**, this will result in redundant traffic to be sent when sending queries, unregistration messages or announcements, as we still sent those to the same multicast group on the same interface.

That's true. However the source address will be different. Suppose we only send out one message for both subnet A and subnet B from a source address in subnet A,  the receiver on subnet B would see it with the source address from a different subnet, which I'm not sure if its valid, or possible at all. I don't think the mdns message is routable.

> I've came up with another approach. We use a tuple(interface index , adress family ) in a set to track already "used" notifications.
> 
> This reduces the number of messages sent drastically for example on my windows machine, i'm getting 3 IPv6 addresses and one IPv4 address. Before this change this would result in 4 messages being send on that interface. With this change we only send 2.

Do you mind paste the `ipconfig` or `ifconfig` or equivalent thing for these interfaces / addresses?

hrzlgnm:
> It's a good question, and I need to dig more as well. Here is my thoughts so far:
> 
> > The socket used to send multicast traffic is acutally bound to the "any" address associated with the respective address family and is joined on the mDNS multicast_v(4|6) address using the network interface index.
> 
> The "bound" part is true, although for "join" it's not using the network interface index for IPv4. It's [using the IP address](https://github.com/keepsimple1/mdns-sd/blob/4772bd199a55da89b49d6e8757995641e4f20b3f/src/service_daemon.rs#L692). IPv6 seems different.

Yes your are right about IPv4, we are using the IP Address to join. But this has the same effect as binding to the interface index from my understanding.

> > If we have multiple IPv6 or IPv4 addresses on the same network interface in different **subnets**, this will result in redundant traffic to be sent when sending queries, unregistration messages or announcements, as we still sent those to the same multicast group on the same interface.
> 
> That's true. However the source address will be different. Suppose we only send out one message for both subnet A and subnet B from a source address in subnet A, the receiver on subnet B would see it with the source address from a different subnet, which I'm not sure if its valid, or possible at all. I don't think the mdns message is routable.
> 

I don't think the subnet matters at all, when receiving message. No-one will look at the senders address and do something with it in regards of mDNS. And yes, according to the standard, the IPv4 mDNS Multicast  address is not routable. [Source](https://en.wikipedia.org/wiki/Multicast_address#Notable_IPv4_multicast_addresses). IPv6 is a different matter.

> > I've came up with another approach. We use a tuple(interface index , adress family ) in a set to track already "used" notifications.
> > This reduces the number of messages sent drastically for example on my windows machine, i'm getting 3 IPv6 addresses and one IPv4 address. Before this change this would result in 4 messages being send on that interface. With this change we only send 2.
> 
> Do you mind paste the `ipconfig` or `ifconfig` or equivalent thing for these interfaces / addresses?

`ip addr show` from my Linux machine, where I'm using a bridge with multicast setup, to be able to do networking stuff from virtual machines:
```
5: mn-bridge1: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue state UP group default qlen 1000
    link/ether 22:06:d3:db:c1:02 brd ff:ff:ff:ff:ff:ff
    inet 192.168.42.92/24 brd 192.168.42.255 scope global noprefixroute mn-bridge1
       valid_lft forever preferred_lft forever
    inet 192.168.43.92/24 brd 192.168.43.255 scope global noprefixroute mn-bridge1
       valid_lft forever preferred_lft forever
    inet 192.168.44.92/24 brd 192.168.44.255 scope global noprefixroute mn-bridge1
       valid_lft forever preferred_lft forever
    inet 192.168.70.92/24 brd 192.168.70.255 scope global noprefixroute mn-bridge1
       valid_lft forever preferred_lft forever
    inet6 fe80::4aa5:3359:1346:2865/64 scope link noprefixroute 
       valid_lft forever preferred_lft forever
```
In this case we would send 5 messages instead of just 2.

Avahi for example randomly uses one of those IPv4 addresses and happily sends queries for example using 192.168.70.92 and is albe to discover services of the other subnets without issues.

`ipconfig` output from the windows machine, sorry about the German:
```
Windows-IP-Konfiguration

Ethernet-Adapter Ethernet:

IPv6-Adresse. . . . . . . . . . . : 2003:e8:bf4e:c900:860d:98d5:d90b:7d 
Temporäre IPv6-Adresse. . . . . . : 2003:e8:bf4e:c900:9532:fd36:933f:ea3d 
Verbindungslokale IPv6-Adresse  . : fe80::f900:996d:50d1:8349%5 
IPv4-Adresse  . . . . . . . . . . : 192.168.178.25 
Subnetzmaske  . . . . . . . . . . : 255.255.255.0

```
In this case we would send 4 messages instead of 2 as mentioned above.

keepsimple1:
> I don't think the subnet matters at all, when receiving message. No-one will look at the senders address and do something with it in regards of mDNS. 

In most cases, I agree it probably doesn't matter. But what if we need to support [Unicast Responses](https://www.rfc-editor.org/rfc/rfc6762#section-5.4) in future?  

And in Known-Answers multi-packet, we also need to know about the source address although we [cannot support it now](https://github.com/keepsimple1/mdns-sd/pull/232#issuecomment-2274958263). 

Of course, one option is to ignore such incoming message if the source is in a different subnet not in the current interface. 

(In general I felt your approach is nice, just wanted to understand all the impacts)

hrzlgnm:
> > I don't think the subnet matters at all, when receiving message. No-one will look at the senders address and do something with it in regards of mDNS.
> 
> In most cases, I agree it probably doesn't matter. But what if we need to support [Unicast Responses](https://www.rfc-editor.org/rfc/rfc6762#section-5.4) in future?

> And in Known-Answers multi-packet, we also need to know about the source address although we [cannot support it now](https://github.com/keepsimple1/mdns-sd/pull/232#issuecomment-2274958263).
> 
> Of course, one option is to ignore such incoming message if the source is in a different subnet not in the current interface.
> 
> (In general I felt your approach is nice, just wanted to understand all the impacts)

Ah ok, I was not aware of those two future use cases. But in general, I think my approach does not put any obstacles in the way of these two points, as its main intent is to reduce the outgoing multicast traffic.

src/service_daemon.rs
```diff
@@ -770,6 +770,18 @@ struct IntfSock {
     sock: Socket,
 }
 
+impl IntfSock {
+    /// Returns the interface index and the associated ip version
+    fn intf_idx_ip_ver(&self) -> (u32, u8) {
```
keepsimple1, line 775: IMO, define a new `struct` for the returned tuple would be good, i.e. improve the readability and rely less on the comments. It will also help in the later parts of the changes.

src/service_daemon.rs
```diff
@@ -770,6 +770,18 @@ struct IntfSock {
...
+        let ip_ver = match self.intf.addr {
+            IfAddr::V4(_) => 4u8,
+            IfAddr::V6(_) => 6u8,
+        };
+
+        (self.intf.index.unwrap_or(0), ip_ver)
```
keepsimple1, line 781: We're using fallback 0 in `set_scope_id`, but in terms of actual network interfaces, it seems their index [starts from 1](https://man7.org/linux/man-pages/man3/if_nameindex.3.html).   We probably want to check the returned index is not 0 to be treated as valid.

src/service_daemon.rs
```diff
@@ -585,8 +585,8 @@ impl ServiceDaemon {
             let now = current_time_millis();
 
             // Notify service listeners about the expired records.
-            let expired_serives = zc.cache.evict_expired_services(now);
-            zc.notify_service_removal(expired_serives);
+            let expired_services = zc.cache.evict_expired_services(now);
+            zc.notify_service_removal(expired_services);
```
keepsimple1, line 589: good catch!

src/service_daemon.rs
```diff
@@ -1199,21 +1211,19 @@ impl Zeroconf {
         self.my_services.insert(service_fullname, info);
     }
 
-    /// Sends out annoucement of `info` on every valid interface.
-    /// Returns the list of interface IPs that sent out the annoucement.
+    /// Sends out announcement of `info` on every valid interface.
+    /// Returns the list of interface IPs that sent out the announcement.
```
keepsimple1, line 1215: good catch!

src/service_daemon.rs
```diff
@@ -1199,21 +1211,19 @@ impl Zeroconf {
...
         let mut outgoing_addrs = Vec::new();
-        let mut subnet_set: HashSet<u128> = HashSet::new();
+        // Send the announcement on one interface per ip version.
+        let mut intf_idx_ip_ver_set: HashSet<(u32, u8)> = HashSet::new();
 
         for (_, intf_sock) in self.intf_socks.iter() {
-            let subnet = ifaddr_subnet(&intf_sock.intf.addr);
-            if !intf_sock.intf.is_link_local() && subnet_set.contains(&subnet) {
-                continue; // no need to send again in the same subnet.
+            if intf_idx_ip_ver_set.contains(&intf_sock.intf_idx_ip_ver()) {
```
keepsimple1, line 1222: Related to a previous comment, I think we probably should check something like:
`if intf_sock.intf_idx_ip_ver().0 != 0`

keepsimple1 (commented):
Overall looks good. A couple of comments inline.

keepsimple1:
> Ah ok, I was not aware of those two future use cases. But in general, I think my approach does not put any obstacles in the way of these two points, as its main intent is to reduce the outgoing multicast traffic.

I think so too. Given that currently we don't specify / limit the source address when browsing services, I think it's worth to go with this optimization.

I submitted the review. And thanks for providing the details about the IP interface and addresses info previously.

src/service_daemon.rs
hrzlgnm: Yes, I agree. I'll make it a struct instead. Perhaps with a method is_valid() that checks just returns sellf.index > 0;

src/service_daemon.rs
hrzlgnm: Yes makes sense, I'll add the checks.

src/service_daemon.rs
```diff
@@ -1,4 +1,4 @@
-//! Service daemon for mDNS Service Discovery.
+///! Service daemon for mDNS Service Discovery.
```
hrzlgnm, line 1: that was not inteded, will revert

src/service_daemon.rs
```diff
@@ -770,6 +770,39 @@ struct IntfSock {
...
+/// Enum to represent the IP version.
+#[derive(Debug, Eq, Hash, PartialEq)]
+enum IpVersion {
+    V4,
+    V6,
+}
+
+/// A struct to track multicast notification status for a network interface.
+#[derive(Debug, Eq, Hash, PartialEq)]
+struct MulticastNotificationTracker {
```
keepsimple1, line 782: I think `IfIndexTracker` (or `MulticastSendTracker`) would be easier to understand / map to what it does. It's up to you to decide though.

src/service_daemon.rs
```diff
@@ -1199,20 +1232,24 @@ impl Zeroconf {
...
         let mut outgoing_addrs = Vec::new();
-        let mut subnet_set: HashSet<u128> = HashSet::new();
+        // Send the announcement on one interface per ip version.
+        let mut multicast_notification_trackers = HashSet::new();
 
         for (_, intf_sock) in self.intf_socks.iter() {
-            let subnet = ifaddr_subnet(&intf_sock.intf.addr);
-            if !intf_sock.intf.is_link_local() && subnet_set.contains(&subnet) {
-                continue; // no need to send again in the same subnet.
+            if intf_sock.as_multicast_notification_tracker().is_some()
```
keepsimple1, line 1243: Wouldn't it be better to extract the value from Option ? like:
```rust
if let Some(tracker) = intf_sock.as_multicast_notification_tracker() {
  if multicast_notification_trackers.contains(tracker) {
     continue;
  }
}
```

src/service_daemon.rs
```diff
@@ -1199,20 +1232,24 @@ impl Zeroconf {
...
+                && multicast_notification_trackers
+                    .contains(&intf_sock.as_multicast_notification_tracker().unwrap())
+            {
+                continue; // No need to send again on the same interface with same ip version.
             }
             if self.broadcast_service_on_intf(info, intf_sock) {
-                if !intf_sock.intf.is_link_local() {
-                    subnet_set.insert(subnet);
+                if intf_sock.as_multicast_notification_tracker().is_some() {
```
keepsimple1, line 1250: same here, how about use `if let Some(tracker) = ...` ?

src/service_daemon.rs
```diff
@@ -770,6 +770,39 @@ struct IntfSock {
...
+/// A struct to track multicast notification status for a network interface.
+#[derive(Debug, Eq, Hash, PartialEq)]
+struct MulticastNotificationTracker {
+    intf_index: u32,
+    ip_version: IpVersion,
+}
+
+impl IntfSock {
+    /// Returns the multicast notification tracker if the interface index is 5valid
+    fn as_multicast_notification_tracker(&self) -> Option<MulticastNotificationTracker> {
```
keepsimple1, line 789: IMO, I don't think this method name needs the `as_` prefix.

src/service_daemon.rs
```diff
@@ -1403,15 +1440,17 @@ impl Zeroconf {
...
-        // Send the query on one interface per subnet.
-        let mut subnet_set: HashSet<u128> = HashSet::new();
+        // Send the query on one interface per ip version.
+        let mut multicast_notification_trackers = HashSet::new();
         for (_, intf_sock) in self.intf_socks.iter() {
-            if !intf_sock.intf.is_link_local() {
-                let subnet = ifaddr_subnet(&intf_sock.intf.addr);
-                if subnet_set.contains(&subnet) {
-                    continue; // no need to send query the same subnet again.
+            if intf_sock.as_multicast_notification_tracker().is_some() {
```
keepsimple1, line 1446: same here for using `if let Some(tracker)`

src/service_daemon.rs
```diff
@@ -2111,7 +2150,19 @@ impl Zeroconf {
             }
             Some((_k, info)) => {
                 let mut timers = Vec::new();
+                // Send one unregister per interface and ip version
+                let mut multicast_notification_trackers = HashSet::new();
+
                 for (ip, intf_sock) in self.intf_socks.iter() {
+                    if intf_sock.as_multicast_notification_tracker().is_some() {
```
keepsimple1, line 2157: same here for using `if let Some(tracker)`

keepsimple1 (commented):
Thanks for the update! A few minor comments inline.

src/service_daemon.rs
hrzlgnm: Went with MulticastSendTracker, thx for the suggestions.

src/service_daemon.rs
hrzlgnm: Ah yes, looks way more like using `rust` :)

src/service_daemon.rs
hrzlgnm: done

src/service_daemon.rs
hrzlgnm: remove the prefix

src/service_daemon.rs
hrzlgnm: done

keepsimple1 (approved):
LGTM, thanks!

PR #271, opened 2024-11-18
feat: replace remaining Box<dyn DnsRecordExt> with type

CosminPerRam:
#206 replaced some remaining `Box<dyn DnsRecordExt>` in members and params, this PR replaces the one in the return position of `DnsRecordExt::clone_box` (and subsequently in each impl), with this there will be only one `Box<>` usage, which is the type definition.

One might uphold that having the full definition in the return position to be better, but I see it as an arbitrary choice as everywhere else we use the type.


keepsimple1 (approved):
Thanks for your PR! LGTM!

PR #279, opened 2024-12-01
update polling to 3.x

Fabian-Gruenbichler:
didn't do any in-depth tests and don't know this code base all to well, so please don't hesitate to do things differently as you see fit.

Debian has recently upgrade to polling 3.x, and this is one of the few packaged crates still using the 2.x release series..

### DIFF
--- a/src/service_daemon.rs
+++ b/src/service_daemon.rs
@@ -43,7 +43,7 @@ use crate::{
 };
 use flume::{bounded, Sender, TrySendError};
 use if_addrs::{IfAddr, Interface};
-use polling::Poller;
+use polling::{Events, Poller};
 use socket2::{SockAddr, Socket};
 use std::{
     cmp::{self, Reverse},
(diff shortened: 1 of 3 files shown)


src/service_daemon.rs
```diff
@@ -482,10 +482,13 @@ impl ServiceDaemon {
     /// 5. process retransmissions if any.
     fn run(mut zc: Zeroconf, receiver: Receiver<Command>) -> Option<Command> {
         // Add the daemon's signal socket to the poller.
-        if let Err(e) = zc.poller.add(
-            &zc.signal_sock,
-            polling::Event::readable(SIGNAL_SOCK_EVENT_KEY),
-        ) {
+        #[allow(unsafe_code)]
```
keepsimple1, line 485: Thanks for your PR! I want to keep this library safe code only, unless we really have to break it. 

Let me take a look at our options.

keepsimple1:
As I wanted to keep this lib safe code only, and `polling` 3.x does not provide a safe API (`add` is unsafe),  I'm trying to move away from `polling` and use `mio` instead (see PR #280 ).  

Could you please confirm if `mio` 1.0 is allowed / packaged in Debian? Thanks!

Fabian-Gruenbichler:
it is (in unstable/testing, which will become trixie/stable next year :))

https://tracker.debian.org/pkg/rust-mio

keepsimple1:
Thanks for confirming. I've merged PR #280, which replaced this PR. Will close this PR now. If any related questions, please feel free to open a new issue or PR. Thanks!

Issue #282, opened 2024-12-09
Feature Request: Extract dns_parser.rs as stand-alone crate. 


rasheedsulayman:
I have a use-case for a tiny DNS client (service-discovery only) for some wasm targets. The problem is that the socket-related libraries (Socket2, if-addrs) used by this crate can't be compiled to those wasm targets yet. i.e Those libraries simply don't have the system call bindings for the targets. As a result of this, I'm just using the the DNS parser utility file from this crate. 

I thought the dns parser utility is useful enough to stand alone as an independent crate.  

An easier alternative would to just make this "mdns-sd" crate expose the dns_parser.rs file publicly, but again, other crates that depend on mdns-sd just for the dns parsing functionality would still need to be able to compile Socket2 and if-addrs. 

So the request to extract dns_parser.rs as stand-alone crate. I'm also available to work on this if needed, but most importantly, I wanted to hear your opinion on it.

keepsimple1:
I didn't realize that the dns_parser file itself is useful for others. Thanks for raising this issue! Usually I'd like to have related things built together with less dependencies, but let me think a bit more about this. And for my curiosity, may I ask what socket/networking crate do you use for wasm targets? (or did you roll your own for sockets?)

rasheedsulayman:
I'm not using any crate. Yes, I'm writing my own unsafe ffi bindings for the socket system calls.

keepsimple1:
Cool! I thought about it and I think it's worth doing it.  I'll give it a shot this weekend and get back to you.

rasheedsulayman:
Thanks.

keepsimple1:
I posted the first version of a PR #284 for this.  I think there are quite a bit more details to iron out, but would you please give it a try (using the PR's branch) to see if it would work for your use case? Any comments are appreciated.

rasheedsulayman:
I have checked. It looks good to me. Just once nitpick comment. 

I was also able to use the newly extracted mdns-parser, so the exposed public APIs are sufficient for my need.

keepsimple1:
Thanks! I've rebased this patch to the latest release 0.13.0 and addressed comments. I will get the diff ready for merging and publishing the new crate, and I plan to do it in a new release specific for this.

rasheedsulayman:
when is the release is likely going to be? i.e when the extracted crate will be available on crate.io.

keepsimple1:
[...]

PR #284, opened 2024-12-15
extract dns_parser.rs into its own crate mdns-parser

keepsimple1:
This patch is trying to resolve issue #282 . At the high level: 

- `dns_parser.rs` is extracted into its own crate `mdns-parser` (not published on crates.io yet) 
- the new crate is kept in this same repo to keep things easier to manage (and the patch diff easier to track).
- `mdns-sd` crate will depend on the new crate `mdns-parser` (inside the repo)


src/lib.rs
```diff
@@ -162,3 +160,6 @@ pub use service_info::{AsIpAddrs, IntoTxtProperties, ServiceInfo, TxtProperties,
 
 /// A handler to receive messages from [ServiceDaemon]. Re-export from `flume` crate.
 pub use flume::Receiver;
+
+/// Re-export to keep backward compatiable.
```
rasheedsulayman, line 164: Nit: Is this comment really needed? I guess It should make no difference to this crate consumers as their code will still work after version upgrade and they won't have any need to reference this comment.

src/dns_cache.rs
```diff
@@ -526,3 +527,11 @@ impl DnsCache {
             .collect()
     }
 }
+
+/// Returns UNIX time in millis
+pub(crate) fn current_time_millis() -> u64 {
```
rasheedsulayman, line 532: Can't we reuse the definition in dns_parser?

rasheedsulayman (approved):

src/dns_cache.rs
rasheedsulayman: On a second thought, I think duplicating is fine since the idea is to the have dns_parser expose only APIs related to parsing.

keepsimple1: 👍  that's exactly why I did that.

src/lib.rs
keepsimple1: You're right. I've updated this comment to be more useful.

PR #286, opened 2024-12-16
fix: service keys must be lowercase

lyager:
Services are registered with fullname as lowercase, when moving waiting_services to active after probe, the lookup failed due to using case service name

### DIFF
--- a/src/service_daemon.rs
+++ b/src/service_daemon.rs
@@ -1468,7 +1468,8 @@ impl Zeroconf {
                     "try to announce service {service_name} on intf {}",
                     intf.ip()
                 );
-                if let Some(info) = self.my_services.get_mut(&service_name) {
+                // service names are lowercase
+                if let Some(info) = self.my_services.get_mut(&service_name.to_lowercase()) {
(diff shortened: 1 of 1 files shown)


lyager:
Maybe the better solution would be to create get/set on the my_services hash?

keepsimple1 (approved):
Thanks for your PR! LGTM!

keepsimple1:
> Maybe the better solution would be to create get/set on the my_services hash?

Not sure about `get/set` on `my_services`, as it's internal data structure. I'm wondering maybe we should use lower case for the keys in the `waiting_services` to be consistent:

https://github.com/keepsimple1/mdns-sd/blob/661408d89ec3d3541345ba5ef6886001324038bc/src/service_info.rs#L807

lyager:
> Not sure about `get/set` on `my_services`, as it's internal data structure. I'm wondering maybe we should use lower case for the keys in the `waiting_services` to be consistent:
> 

Consistency is always good :) There's also UniCase, but it might be going overboard. Thanks for the quick feedback.

Issue #299, opened 2025-02-04
Consume `ServiceInfo` data building `ServiceInfoData` structure


Luni-4:
Currently, I'm facing a specific workflow which needs to consume `ServiceInfo`. In my opinion, it would be helpful to add a `ServiceInfoData` structure containing all information that can be retrieved through `get_*` functions.

Something like:

```rust
impl ServiceInfo {
  pub fn service_info_data(self) -> ServiceInfoData {
      ServiceInfoData {
             fullname: self.fullname,
             addresses: self.addresses,
             port: self.port,
             properties: self.properties.properties_data(),
             ...
      }
  }
}
```
The example above is not exhaustive, therefore some public fields are missing. This approach would reduce the amount of allocations to be performed when `ServiceInfo` must not be saved.

keepsimple1:
You have a good point. I think the problem is that currently the same `ServiceInfo` struct is used in both the server and the client, i.e. in publishing a service ([`register`](https://docs.rs/mdns-sd/0.13.2/mdns_sd/struct.ServiceDaemon.html#method.register))  and in resolving a service instance ([`ServiceEvent::ServiceResolved`](https://docs.rs/mdns-sd/0.13.2/mdns_sd/enum.ServiceEvent.html)) 

We probably should have defined a different struct for the client side in resolving a service, that contains the attributes as in `ServiceInfoData` you mentioned.  Assuming you intended to use `ServiceInfoData` only on the client (browser) side, I think there are a couple of options going forward: 

1. Like you suggested, add a new method in `ServiceInfo` to return a new struct.  The only thing is that I felt the name `ServiceInfoData` could cause confusions, as `data` is quite generic and similar to `info`. 

2. Replace the struct in `ServiceResolved(ServiceInfo)` with a new struct that is smaller than `ServiceInfo` and it's clearly only used on the client side.  The downside is that it will be a breaking change, even if we provide all same `getter` functions  of `ServiceInfo`.
 
In both options, how about naming the new struct `ResolvedService` ?  It clearly indicates the purpose of the struct, and will only used by the client side. 

What do you think?

Luni-4:
> You have a good point. Do you intend to use `ServiceInfoData` only on the client (browser) side?

Yep, exactly, only on the client side.

[...]

PR #302, opened 2025-02-06
ResolvedService: a new plain struct for attributes of a service

keepsimple1:
This patch is to follow up on the discussions in issue #299.

### DIFF
--- a/src/service_info.rs
+++ b/src/service_info.rs
@@ -1103,8 +1103,6 @@ pub(crate) fn split_sub_domain(domain: &str) -> (&str, Option<&str>) {
 #[non_exhaustive]
 pub struct ResolvedService {
     /// service type and domain. For example, "_http._tcp.local."
-    /// Note: when subtype is used, ty_domain might not be a suffix
-    /// of fullname.
     pub ty_domain: String,
 
     /// full name of the service. For example, "my-service._http._tcp.local."
--- a/src/service_info.rs
+++ b/src/service_info.rs
@@ -367,7 +367,7 @@ impl ServiceInfo {
             .unwrap_or(ServiceStatus::Unknown)
     }
 
-    /// Returns a resolved service instance.
+    /// Consumes self and returns a resolved service, i.e. a lite version of `ServiceInfo`.
     pub fn as_resolved_service(self) -> ResolvedService {
         ResolvedService {
             ty_domain: self.ty_domain,
@@ -1102,11 +1102,24 @@ pub(crate) fn split_sub_domain(domain: &str) -> (&str, Option<&str>) {
 /// This is from a client (i.e. querier) point of view.
 #[non_exhaustive]
 pub struct ResolvedService {
+    /// service type and domain. For example, "_http._tcp.local."
+    /// Note: when subtype is used, ty_domain might not be a suffix
+    /// of fullname.
     pub ty_domain: String,
+
+    /// full name of the service. For example, "my-service._http._tcp.local."
     pub fullname: String,
+
+    /// host name of the service. For example, "my-server1.local."
     pub host: String,
+
+    /// port of the service. I.e. TCP or UDP port.
     pub port: u16,
+
+    /// addresses of the service. IPv4 or IPv6 addresses.
     pub addresses: HashSet<IpAddr>,
+
+    /// properties of the service, decoded from TXT record.
     pub txt_properties: TxtProperties,
 }
 
--- a/src/service_info.rs
+++ b/src/service_info.rs
@@ -366,6 +366,18 @@ impl ServiceInfo {
             .cloned()
             .unwrap_or(ServiceStatus::Unknown)
     }
+
+    /// Returns a resolved service instance.
+    pub fn as_resolved_service(self) -> ResolvedService {
+        ResolvedService {
+            ty_domain: self.ty_domain,
+            fullname: self.fullname,
+            host: self.server,
+            port: self.port,
+            addresses: self.addresses,
+            txt_properties: self.txt_properties,
+        }
+    }
 }
 
 /// Removes potentially duplicated ".local." at the end of "hostname".
@@ -1084,6 +1096,16 @@ pub(crate) fn split_sub_domain(domain: &str) -> (&str, Option<&str>) {
     } else {
         (domain, None)
     }
+}
+
+#[non_exhaustive]
+pub struct ResolvedService {
+    pub ty_domain: String,
+    pub fullname: String,
+    pub host: String,
+    pub port: u16,
+    pub addresses: HashSet<IpAddr>,
+    pub txt_properties: TxtProperties,
 }
 
 #[cfg(test)]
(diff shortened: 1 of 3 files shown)


src/service_info.rs
```diff
@@ -1060,6 +1072,16 @@ pub(crate) fn split_sub_domain(domain: &str) -> (&str, Option<&str>) {
     }
 }
 
+#[non_exhaustive]
+pub struct ResolvedService {
```
keepsimple1, line 1076: One edge case is:  if `addresses` is empty in this struct, then it would not be `resolved`.  Should we call this struct just `Service` instead?

src/service_info.rs
```diff
@@ -366,6 +366,18 @@ impl ServiceInfo {
...
+
+    /// Returns a resolved service instance.
+    pub fn as_resolved_service(&self) -> ResolvedService {
+        ResolvedService {
+            ty_domain: self.ty_domain.clone(),
+            fullname: self.fullname.clone(),
+            host: self.server.clone(),
+            port: self.port,
+            addresses: self.addresses.clone(),
+            txt_properties: self.txt_properties.clone(),
```
keepsimple1, line 378: To be consistent with `ServiceInfo`, I am cloning `TxtProperties`,  not creating a new `HashMap` of entries inside `TxtProperties` (see issue #299).  We can handle data extraction of `TxtProperties` separately.

src/service_info.rs
```diff
@@ -366,6 +366,18 @@ impl ServiceInfo {
             .cloned()
             .unwrap_or(ServiceStatus::Unknown)
     }
+
+    /// Returns a resolved service instance.
+    pub fn as_resolved_service(&self) -> ResolvedService {
```
keepsimple1, line 371: If we name the struct as `Service`, this function will be `as_service`.

src/service_info.rs
Luni-4: If we have decided to split among client and server, we can return an `Option` like we are doing right now, indeed Option<ServiceInfo>` can have empty `addresses`. Therefore, `Option<ResolvedService>` could it be a solution?

Luni-4:
This approach does not consume `ServiceInfo`, just clones its internal fields, so you have two structs in memory at the same time

keepsimple1:
> This approach does not consume ServiceInfo, just clones its internal fields, so you have two structs in memory at the same time

I made the mistake in the 1st commit, please check the latest one. ;-)

Luni-4 (approved):
This is fine for me, thanks a lot! :)

Just a simple nit

src/service_info.rs
```diff
@@ -366,6 +366,18 @@ impl ServiceInfo {
             .cloned()
             .unwrap_or(ServiceStatus::Unknown)
     }
+
+    /// Returns a resolved service instance.
+    pub fn as_resolved_service(self) -> ResolvedService {
```
Luni-4, line 371: Nit: Do we use `Option<ResolvedService>` for empty addresses?

keepsimple1: I have thought about this, and wanted to keep returning just `ResolvedService` for a couple of reasons: 

1. Better ergonomics for the user.  (Having `Option` can complicate many things for the user, in my experience).
2. Anyway `ResolvedService` struct cannot guarantee its fields are all valid, for example what if `.fullname`  is empty?  So I think we can implement methods in `ResolvedService` to valid its content, something like `is_valid()`, etc.

Luni-4: I agree with your points, but I'm a bit confused. 

Currently, we retrieve a service client side in this way (just an example with timeout): 

```
while let Ok(event) = receiver.recv_timeout(Duration::from_secs(1)) {
        if let ServiceEvent::ServiceResolved(info) = event {
            info!("{:?}", info);
        }
    }
```

Now we are going to replace `info` variable, which is a `ServiceInfo` structure, with `ResolvedService`. But if there isn't a valid `fullname` or an empty vector of `addresses`, can't we block all of this before at the `ServiceEvent::ServiceResolved(info)` condition? I mean, we can return a different `event` variant when the conditions above are not satisfied. Just talking for the client side.

keepsimple1: > Now we are going to replace info variable, which is a ServiceInfo structure, with ResolvedService. 

Actually I'm not saying to replace `info` here.  Because such change will break the backward compatibility, we said going with option 1 instead:  The user will call `info.as_resolved_service` to consume `info` and get a `ResolvedService` struct.

> But if there isn't a valid fullname or an empty vector of addresses, can't we block all of this before, when we are evaluating the ServiceEvent::ServiceResolved(info) condition? 

Yes, from this event the `info` is always valid, which means `ResolvedService` struct will also be valid in this context.  But this is a business logic, we can document it but we cannot enforce it (via the type system or compiler). 

But for any given `ResolvedService` without such context, we would need something like `is_valid` (or `is_ready` to match the current code in service_info.rs).

Luni-4: Ok, I see now, thank you! The changes are fine for me

src/service_info.rs
```diff
@@ -366,6 +366,18 @@ impl ServiceInfo {
             .cloned()
             .unwrap_or(ServiceStatus::Unknown)
     }
+
+    /// Consumes self and returns a resolved service, i.e. a lite version of `ServiceInfo`.
+    pub fn as_resolved_service(self) -> ResolvedService {
+        ResolvedService {
+            ty_domain: self.ty_domain,
```
keepsimple1, line 373: Another question:  what if this service info has some `sub_domain` ?  Currently, `ServiceInfo` stores only the regular service type & domain in `ty_domain`, and the full subtype & domain in `sub_domain`.   Do we want the same thing in `ResolvedService`?

src/service_info.rs
keepsimple1: Thanks!  I've updated the diff with a test case and export in `lib.rs`.  However, I have a new question regarding `sub_domain` below.  Any suggestions or comments?

src/service_info.rs
Luni-4: I think we should make `sub_domain` public too. If we are consuming `ServingInfo`, we can move that information into `ResolvedService`. We do not know the kind of use cases out there, perhaps someone needs that information

keepsimple1: `ty_domain` and `sub_domain` was not public, but they are indeed accessible via a public API (getter). I agree with you, I think it's better to include `sub_domain` in `ResolvedService` as well.

Luni-4: With public, I meant accessible via public interface :P

keepsimple1: I've updated the diff and added `sub_ty_domain`. No longer a draft ;-).  Let me know if this patch works for you. Thanks!

src/service_info.rs
```diff
@@ -1086,6 +1099,47 @@ pub(crate) fn split_sub_domain(domain: &str) -> (&str, Option<&str>) {
     }
 }
 
+/// Represents a resolved service as a plain data struct.
+/// This is from a client (i.e. querier) point of view.
+#[non_exhaustive]
+pub struct ResolvedService {
+    /// service type and domain. For example, "_http._tcp.local."
```
Luni-4, line 1106: Nit: Shouldn't the comments of this structure capitalized? The `sub_ty_domain` is capitalized and even the functions ones

keepsimple1: Fixed. Thanks!

Issue #300, opened 2025-02-04
Use `indexmap` for `TxtProperties`


Luni-4:
I was just looking at these [code lines](https://github.com/keepsimple1/mdns-sd/blob/main/src/service_info.rs#L456-L457) and I've thought that the [indexmap](https://crates.io/crates/indexmap) collection would be efficient for `TxtProperties`'s use-cases.

Advantages:

 - Efficient because a `TxtProperty` **cannot** be removed from the collection, but only read. This would **always** respect the insertion order. I've based my assumption from the methods provided by this structure though
 - Fast iteration
 - On par with `HashMap` in terms of performance

Disadvantages:

- A new dependency
- The dependency might be unmaintained one day

It would also be helpful to provide a method to access the internal `indexmap`

keepsimple1:
Something like `indexmap` would be indeed more efficient than `Vec<>` used today. However, I think in this particular case, the downside (as you also mentioned) overweights the upside.  One reason is that `TxtProperties` mostly has just a few items, hence the difference between using `Vec` and `indexmap` would be small.  On the other hand, fewer dependencies is a good thing to keep.

If any real world use case (or measurements) shows that `TxtProperties` is a performance bottleneck, then we can look into it more.

Luni-4:
I agree with your analysis. The number of `TxtProperty` is usually low, hence the performance impact remains small independently of the collection in use.

As personal use-case, I need to convert `TxtProperties` into `HashMap<String, String>`, so the use of `IndexMap` would have been simpler. As it is now, we need to allocate `String`s from `&str` returned by the `val_str()` method. Perhaps, adding a method which consumes `TxtProperties` into an `HashMap<String, String>` would be the most suitable solution for my use-case.

keepsimple1:
> Perhaps, adding a method which consumes TxtProperties into an HashMap<String, String> would be the most suitable solution for my use-case.

I think we can add a new method for such cases.  I've opened a draft PR #303  based on my understanding. Please let me know if that would work for you.

keepsimple1:
PR merged. Thanks!

PR #303, opened 2025-02-06
TxtProperties: new method to get a HashMap of properties

keepsimple1:
This is to follow up on issue #300 . Added a new method that consumes `TxtProperties` and returns a simple HashMap.

### DIFF
--- a/src/service_info.rs
+++ b/src/service_info.rs
@@ -501,12 +501,17 @@ impl TxtProperties {
     }
 
     /// Returns a map of properties, where the key is the property key.
-    pub fn get_property_map_str(&self) -> HashMap<String, String> {
-        let mut map = HashMap::new();
-        for prop in self.properties.iter() {
-            map.insert(prop.key.clone(), prop.val_str().to_string());
-        }
-        map
+    pub fn into_property_map_str(self) -> HashMap<String, String> {
+        self.properties
+            .into_iter()
+            .map(|p| {
+                (
+                    p.key,
+                    p.val
+                        .map_or_else(String::new, |v| String::from_utf8(v).unwrap_or_default()),
+                )
+            })
+            .collect()
     }
 }
 
--- a/src/service_info.rs
+++ b/src/service_info.rs
@@ -499,6 +499,15 @@ impl TxtProperties {
     pub fn get_property_val_str(&self, key: &str) -> Option<&str> {
         self.get(key).map(|x| x.val_str())
     }
+
+    /// Returns a map of properties, where the key is the property key.
+    pub fn get_property_map_str(&self) -> HashMap<String, String> {
+        let mut map = HashMap::new();
+        for prop in self.properties.iter() {
+            map.insert(prop.key.clone(), prop.val_str().to_string());
+        }
+        map
+    }
 }
 
 impl fmt::Display for TxtProperties {
--- a/src/service_info.rs
+++ b/src/service_info.rs
@@ -500,16 +500,21 @@ impl TxtProperties {
         self.get(key).map(|x| x.val_str())
     }
 
-    /// Returns a map of properties, where the key is the property key.
+    /// Returns properties as a hashmap, where the keys are the properties keys.
     pub fn into_property_map_str(self) -> HashMap<String, String> {
         self.properties
             .into_iter()
-            .filter_map(|p| {
-                // Skip the property if the value is non-empty and not valid UTF-8.
-                let value = p
-                    .val
-                    .map_or(Some(String::new()), |v| String::from_utf8(v).ok())?;
-                Some((p.key, value))
+            .filter_map(|property| {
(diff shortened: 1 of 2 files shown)


src/service_info.rs
```diff
@@ -499,6 +499,15 @@ impl TxtProperties {
     pub fn get_property_val_str(&self, key: &str) -> Option<&str> {
         self.get(key).map(|x| x.val_str())
     }
+
+    /// Returns a map of properties, where the key is the property key.
+    pub fn get_property_map_str(&self) -> HashMap<String, String> {
```
Luni-4: Can we consume the `TxtProperties`? Something like:

```rust
/// Returns a map of properties, where the key is the property key.
///
/// **This method consumes [`TxtProperties`].**
pub fn into_hashmap(self) -> HashMap<String, String>
```

One can always obtain an `HasMap<&str, &str>` using the current methods provided by `TxtProperties`

src/service_info.rs
```diff
@@ -499,6 +499,15 @@ impl TxtProperties {
...
+        let mut map = HashMap::new();
+        for prop in self.properties.iter() {
```
Luni-4: Connecting to the comment above, I would use `into_iter()` here

src/service_info.rs
```diff
@@ -499,6 +499,15 @@ impl TxtProperties {
...
+            map.insert(prop.key.clone(), prop.val_str().to_string());
```
Luni-4: If we consume, we can use `prop` directly

Luni-4 (commented):
Thanks a lot for your PR, very appreciated! :)

Some changes I would make, let me know if they are feasible in your opinion

src/service_info.rs
keepsimple1: There are 2 commits. The 2nd commit already consumes the `TxtProperties`.

src/service_info.rs
keepsimple1: same here, it's in the 2nd commit ;-)

src/service_info.rs
keepsimple1: same here, it's in the 2nd commit.

src/service_info.rs
```diff
@@ -501,12 +501,17 @@ impl TxtProperties {
...
-        }
-        map
+    pub fn into_property_map_str(self) -> HashMap<String, String> {
+        self.properties
+            .into_iter()
+            .map(|p| {
+                (
+                    p.key,
+                    p.val
+                        .map_or_else(String::new, |v| String::from_utf8(v).unwrap_or_default()),
```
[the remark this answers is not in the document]
keepsimple1: Sounds good! Will update.

src/service_info.rs
```diff
@@ -499,6 +499,20 @@ impl TxtProperties {
     pub fn get_property_val_str(&self, key: &str) -> Option<&str> {
         self.get(key).map(|x| x.val_str())
     }
+
+    /// Returns a map of properties, where the key is the property key.
+    pub fn into_property_map_str(self) -> HashMap<String, String> {
+        self.properties
+            .into_iter()
+            .filter_map(|p| {
+                // Skip the property if the value is non-empty and not valid UTF-8.
```
Luni-4, line 508: The comment seems wrong. We should skip if the value is **empty** and not valid UTF-8

src/service_info.rs
```diff
@@ -499,6 +499,20 @@ impl TxtProperties {
...
+                let value = p
+                    .val
+                    .map_or(Some(String::new()), |v| String::from_utf8(v).ok())?;
```
Luni-4, line 511: I would use [is_some_and](https://doc.rust-lang.org/std/option/enum.Option.html#method.is_some_and) instead of `map_or`, in this way we do not consider empty `String`s

Luni-4 (changes requested):
Last changes and it is fine for me

src/service_info.rs
keepsimple1: I indeed meant **non-empty**.

As  TXT record allows empty value for a key (see [RFC](https://datatracker.ietf.org/doc/html/rfc6763#section-6.4), I wanted to do:

- If the value is empty, this function sets an empty string for this attribute.
- If the value is non-empty but fails to convert to UTF-8, this function treats it as invalid, and skip this attribute. (Maybe we should log a debug message).

src/service_info.rs
keepsimple1: related to the previous comment resolution.

src/service_info.rs
Luni-4: I did not know that, clear now, thank you for your explanation! :)

Luni-4 (approved):
This is fine for me, thank you!

keepsimple1:
Thank you for your review! I updated the comments to make it clearer and added a test case.

Luni-4:
Cool! Thanks a lot for your update!

PR #304, opened 2025-02-06
Remind users of the maximum length of the attribute.

mycrl:


src/service_info.rs
```diff
@@ -691,6 +694,11 @@ fn encode_txt<'a>(properties: impl Iterator<Item = &'a TxtProperty>) -> Vec<u8>
             s.extend(v);
         }
 
+        assert!(
```
keepsimple1, line 697: as this is a library, I would avoid `asset!`. Instead, we can ignore the property or truncate the `s` if it's too long, and log a debug message.

keepsimple1 (commented):
Thanks for your PR! A couple of comments inline.

src/service_info.rs
mycrl: Direct truncation would be better, it helps the user to notice that the attribute is truncated and draw attention to it, dropping it would confuse the user.

keepsimple1 (commented):
Thanks for your update!  Could you please add a test?

mycrl:
Added a simple text test. Considering the limited impact of this feature on functionality, no special treatment is required.

keepsimple1 (approved):
LGTM!  Thanks!

PR #306, opened 2025-02-12
fix: include loopback addresses in filtering criteria

minetake01:
Hi,  

This PR updates the filtering criteria to include loopback addresses when returning a list of addresses in the same LAN.  

In some use cases, particularly with OSCQuery, it may be necessary to publish loopback addresses via mDNS. Allowing loopback addresses ensures better compatibility with such scenarios.  

Let me know if any changes are needed. Thanks!

### DIFF
--- a/src/service_info.rs
+++ b/src/service_info.rs
@@ -291,6 +291,8 @@ impl ServiceInfo {
     pub(crate) fn get_addrs_on_intf(&self, intf: &Interface) -> Vec<IpAddr> {
         self.addresses
             .iter()
+            // Allow loopback addresses to support registering services on loopback interfaces,
+            // which is required by some use cases (e.g., OSCQuery) that publish via mDNS.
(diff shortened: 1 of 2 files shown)


src/service_info.rs
```diff
@@ -291,7 +291,7 @@ impl ServiceInfo {
     pub(crate) fn get_addrs_on_intf(&self, intf: &Interface) -> Vec<IpAddr> {
         self.addresses
             .iter()
-            .filter(|a| valid_ip_on_intf(a, intf))
+            .filter(|a| (a.is_loopback() || valid_ip_on_intf(a, intf)))
```
keepsimple1, line 294: I assume this is to support registering a service with a loopback address (maybe in addition to other addresses)? Seem OK to me.  Could you please add a comment here, and a test case? 

And I'm also curious why it's needed: currently this lib can already find local responders via the (non-loopback) IP address.

keepsimple1 (approved):
Thanks for your update!  Looks good!

PR #307, opened 2025-02-14
Including Loopback Addresses Properly and Proposing a Feature Flag

minetake01:
Hello,

This pull request addresses an issue with how we handle network interface enumeration. In a previous PR (#306 ), we modified the behavior to allow returning loopback addresses. However, because announcements are not sent over the loopback address space, other services have had difficulty discovering published services.

To resolve the discovery issue, this change removes the exclusion filter so that loopback addresses are included. However, note that announcing on loopback interfaces may introduce potential performance issues, since announcements would then always be made on these interfaces.

Proposal:
I suggest that we consider reverting the previous PR (#306) and apply this change behind a feature flag or configuration option. This would allow users to opt in or out of loopback announcements based on their performance needs and use case.

Please review the changes and share any feedback or concerns regarding the proposed feature flag approach. I appreciate your time and look forward to your suggestions.

Thank you.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,6 +14,7 @@ description = "mDNS Service Discovery library with no async runtime dependency"
 [features]
 async = ["flume/async"]
 logging = ["log"]
+loopback = []
 default = ["async", "logging"]
 
 [dependencies]
--- a/src/service_daemon.rs
+++ b/src/service_daemon.rs
@@ -3121,6 +3121,18 @@ fn call_hostname_resolution_listener(
 }
 
 /// Returns valid network interfaces in the host system.
+/// Loopback interfaces are excluded.
+#[cfg(not(feature = "loopback"))]
+fn my_ip_interfaces() -> Vec<Interface> {
+    if_addrs::get_if_addrs()
+        .unwrap_or_default()
+        .into_iter()
+        .filter(|i| !i.is_loopback())
+        .collect()
+}
+
+/// Returns valid network interfaces in the host system.
+#[cfg(feature = "loopback")]
 fn my_ip_interfaces() -> Vec<Interface> {
     if_addrs::get_if_addrs()
         .unwrap_or_default()
--- a/src/service_daemon.rs
+++ b/src/service_daemon.rs
@@ -3121,12 +3121,10 @@ fn call_hostname_resolution_listener(
 }
 
 /// Returns valid network interfaces in the host system.
-/// Loopback interfaces are excluded.
 fn my_ip_interfaces() -> Vec<Interface> {
     if_addrs::get_if_addrs()
         .unwrap_or_default()
         .into_iter()
-        .filter(|i| !i.is_loopback())
         .collect()
 }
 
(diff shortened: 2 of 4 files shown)


minetake01:
Sorry, after applying the changes to my crate, I encountered an issue where service registration gets stuck. I'm currently investigating the cause. I'll let you know once I've fixed it and written test cases.

minetake01:
Sorry, I applied the changes to my crate and found that service name resolution for loopback addresses no longer works. It could be a temporary issue due to caching or other factors, but I’m still testing it. I’ll update you once I identify the cause.

keepsimple1:
While you are working on the issue, I wanted to say that PR #306 was OK as it only lists the loopback address as part of `ServiceInfo` addresses, without changing the interfaces sending the announcement packets. It's a different story if we want to announce the service via the loopback interface.

I think the current codebase already supports your use case (as following). But I could be wrong. 
1. Publishing a service on localhost via non-loopback IP.  (say 192.168.0.100), including loopback (127.0.0.1) in its addresses.
2. Starts a client querying the service.
3. The client receives mDNS packets from the IP 192.168.0.100.  The `ServiceInfo.get_addresses()` returns a list including the loopback address.
4.  The client can choose to visit the application server endpoint (say OSC) using the loopback address. 

Please comment on what I'm missing above to help me understand the rationale of this PR.

minetake01:
Thanks for your reply.
After thinking it over, I realized that you’re right about this. I apologize for the confusion. I’ll go ahead and close this PR.

keepsimple1:
No worries ;-)

Also I was thinking about this again today. I am still curious: was it necessary to have `127.0.0.1` (loopback addr) in the address list of `ServiceInfo` ?  Is it because the host was offline, i.e. might not have any other IP address?  Like I mentioned earlier, if the host has any other regular IP (say 192.168.0.100), the client can find it via the regular IP and communicate with it.

Or, is it because the loopback address is stable and easy to use,  while the regular IP might change and is unpredictable? (probably even  `ServiceInfo.enable_addr_auto()` would not work as good as the loopback address?)

minetake01:
mDNS is generally not intended for use on loopback addresses, but in VRChat’s OSCQuery implementation, mDNS is used to discover tools running on the same host. Since these tools do not necessarily bind to a non-loopback address, the loopback address must be included in `ServiceInfo` for discovery to work properly.

However, as far as I know, this kind of use case is unique to VRChat, so whether this change should be included in the next release is still up for discussion.

keepsimple1:
> However, as far as I know, this kind of use case is unique to VRChat, so whether this change should be included in the next release is still up for discussion.

Thanks for the details. I understand better now. Indeed there is an issue I missed earlier: 

With the current code, when a host (A) publishes `ServiceInfo` with the loopback address (127.0.0.1),  other hosts  (B) on the same network also sees this address.  This is obviously wrong for host B.

I think we would need to properly support the loopback interface. I will take a look.

keepsimple1:
I've opened a new PR #317 to properly support the loopback interface. It passed the existing test of `service_with_loopback_addr` (by calling `enable_interface`).  Please check if you have time. Let me know what you think. Thanks!

minetake01:
Thanks! I'll check it out later.

keepsimple1:
I plan to merge the PR #317 to prepare for a new release soon.  But don't worry, you can always give it a try when you have time. Thanks!

PR #310, opened 2025-02-15
ci: add doc check and fail if there are any warnings

CosminPerRam:


keepsimple1 (approved):
LGTM! Thanks for adding this!

PR #316, opened 2025-02-23
fix: remove related addresses in the cache when disabling an interface

keepsimple1:
This is a follow up fix of PR #296, and in turn issue #295 .

There are two changes:
1.  there is a timing window between starting a daemon and disabling an interface. The cache could already have stored some addresses before an interface is disabled. To make sure such address will be not resolved, we need to remove them when the interface is disabled.

2. On Linux, tests show that when multicast packets loop back, they are received by all interfaces that joined the multicast group, not just the original interface that sent the packet. As the result, interfaces of different network would also get the packet.  This behavior impact this PR as both server and client are running on the same host, and all multicast packets loop back (by default). Hence added a check against such case.

### DIFF
--- a/tests/mdns_test.rs
+++ b/tests/mdns_test.rs
@@ -681,6 +681,81 @@ fn service_with_ipv4_only() {
 }
 
 #[test]
+fn test_disable_interface_cache() {
+    // Create a server
+    let server = ServiceDaemon::new().expect("Failed to create the server");
+
+    // Register a service with one IPv4.
+    let ty_domain = "_disable-intf._tcp.local.";
+    let now = SystemTime::now()
+        .duration_since(SystemTime::UNIX_EPOCH)
+        .unwrap();
+    let instance_name = now.as_micros().to_string();
+    let service_ip_addr = my_ip_interfaces()
+        .iter()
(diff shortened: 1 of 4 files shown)


tests/mdns_test.rs
```diff
@@ -680,6 +680,81 @@ fn service_with_ipv4_only() {
...
+    // Disable the interface for the client.
+    client.disable_interface(service_ip_addr).unwrap();
+
+    // Browse for the service.
+    let handle = client.browse(&ty_domain).unwrap();
+    let timeout = Duration::from_secs(1);
+    let mut resolved = false;
+
+    // run till timeout and it should not resolve.
+    loop {
```
Luni-4, line 730: Out of curiosity, what's the main purpose of this loop? I cannot figure it out. Is there the possibility that `recv_timeout` does not close after the `timeout` is expired?

keepsimple1: `loop` is needed because `recv_timeout` could first return events other than `ServiceResolved` ( for example `ServiceFound`) , but we have to wait for `ServiceResolved` or a timeout error.

Luni-4: Ah, I see now, thanks for your explanation!

PR #323, opened 2025-03-06
feat: add boxed in DnsRecordExt to reduce Box::new usage

CosminPerRam:
This is the equivalent of `DnsRecordExt::clone_box` but consuming.


keepsimple1 (approved):
LGTM!

PR #324, opened 2025-03-07
chore(deps): update fastrand and socket2

CosminPerRam:
I'd like for this to be merged mainly due to `socket2` updating `windows-sys` from `0.48` to `0.52`, thus getting rid of a repeated dependency in my project (due to other packages using the newer version).

This PR also updates `fastrand` in dev-dependencies ~~and removes 'test-log-macros' as it is unused.~~ (edit: not yet)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,16 +17,15 @@ logging = ["log"]
 default = ["async", "logging"]
 
 [dependencies]
-fastrand = "2.1"
+fastrand = "2.3"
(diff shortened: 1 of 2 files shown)


keepsimple1:
Thanks for your PR. Looks like CI failed due to `test-log-macros` problem. Maybe keep the version of `test-log-macros` as we are not upgrading to rustc 1.71 yet.

keepsimple1 (approved):
Thanks for your update! LGTM!

Issue #322, opened 2025-03-05
Occasional failure of service discovery daemon to detect IP address coming up


hgomersall:
During soak testing, I have observed a rare problem in which a new IP address that becomes active after the service discovery daemon begins is not picked up as expected by the daemon.

The sequence is as follows:

1. System boots
2. Software using mdns-sd comes up
3. ~50s later link-local IP address is assigned to the interface
4. 20-30s later the service is normally advertised

The problem is step 4 occasionally (<1 in 200 times) does not work despite the IP address coming up properly as expected.

The tracing output sits at the following and never moves on (which it does when it works properly):
```
2025-01-15T16:16:24.437700Z TRACE mio::poll: reregistering event source with poller: token=Token(1), interests=READABLE
```
This is using Linux on Arm7.

keepsimple1:
I was able to reproduce this problem and found the cause (see PR #328 ) . Thanks again for reporting this issue with details!

hgomersall:
Out of interest, how did you go about reproducing this? I ask in case there is something better I can be doing at my end for these sorts of tests.

keepsimple1:
I kind of did the simplest thing ;-):   created a Docker container of Ubuntu Linux, and run a shell script to follow your use case. Here is the script on gist:  https://gist.github.com/keepsimple1/684f1cf337e0a5057fbc4f1c954961ed.  (I added this info in the fix PR)

PR #328, opened 2025-03-19
bugfix: when now equals next_ip_check

keepsimple1:
This is to fix issue #322 . I reproduced the issue locally and found that it would happen if `now` equals `next_ip_check` and skips the IP changes check. Hence the diff.  I no longer see the issue with the fix.


hgomersall:
Great work, thanks for the fix!

PR #332, opened 2025-03-25
bugfix: make `get_addr` method use lowercase hostname

Stupremee:
In the last pull request regarding the lowercase hostname topic, the resolution of services was broken.

The version `0.13.4` should probably be yanked, to avoid updates where suddenly resolving of services is broken.

I have also adjusted the integration test, so it would have catched this case.


hrzlgnm:
I've made an issue for the same topic #333

keepsimple1 (approved):
Sorry about the breakage, thanks for your fix! LGTM.

Issue #308, opened 2025-02-15
test test_multicast_loop_v6 failed


CosminPerRam:
Whilst contributing to the crate in the past, sometimes when running the tests some would fail randomly quite commonly, so I wanted to check if this is still the case now.

Running the tests 10 times on MacOS Sequoia 15.2 had no fails.
Running the tests 10 times on Windows 11 Pro 23H2 22631.4751 yields in `test_multicast_loop_v6` failure on the 2nd run, running that specific test 10 more times with no fails.

Failed test logs:
```
[2025-02-15T14:49:03Z DEBUG mdns_sd::service_daemon] register service ServiceInfo { ty_domain: "_loop_v6._udp.local.", sub_domain: None, fullname: "1739630943026520._loop_v6._udp.local.", server: "loop_v6_host.local.", addresses: {fe80::2646:9c71:3cff:e9cc}, port: 5200, host_ttl: 120, other_ttl: 4500, priority: 0, weight: 0, txt_properties: TxtProperties { properties: [] }, addr_auto: false, status: {}, requires_probe: true }
[2025-02-15T14:49:03Z DEBUG mdns_sd::service_daemon] prepare to announce service 1739630943026520._loop_v6._udp.local. on vEthernet (Default Switch): fe80::2646:9c71:3cff:e9cc
[2025-02-15T14:49:03Z DEBUG mdns_sd::service_info] new probe of 1739630943026520._loop_v6._udp.local.
[2025-02-15T14:49:03Z DEBUG mdns_sd::service_info] insert record TYPE_SRV into probe of 1739630943026520._loop_v6._udp.local.
[2025-02-15T14:49:03Z DEBUG mdns_sd::service_info] insert record TYPE_TXT into probe of 1739630943026520._loop_v6._udp.local.
[2025-02-15T14:49:03Z DEBUG mdns_sd::service_info] new probe of loop_v6_host.local.
[2025-02-15T14:49:03Z DEBUG mdns_sd::service_info] insert record TYPE_AAAA into probe of loop_v6_host.local.
[2025-02-15T14:49:03Z DEBUG mdns_sd::service_daemon] prepare to announce service 1739630943026520._loop_v6._udp.local. on vEthernet (WSL (Hyper-V firewall)): fe80::1c07:dab:2037:fb7f
[2025-02-15T14:49:03Z DEBUG mdns_sd::service_info] new probe of 1739630943026520._loop_v6._udp.local.
[...]

PR #335, opened 2025-03-29
ServiceDaemon: new method to clear cached DNS records

keepsimple1:
This patch is trying to resolve issue #308 .  One thing I'm not sure is that this lib did not expose the concept of `cache` before. I'm wondering if `clear_cache()` should be called something else so that existing users can understand more easily.

### DIFF
--- a/src/dns_cache.rs
+++ b/src/dns_cache.rs
@@ -576,6 +576,16 @@ impl DnsCache {
             });
         }
     }
+
+    /// Flush the cache for all records.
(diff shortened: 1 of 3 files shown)


keepsimple1:
I think PR #338 also addresses this issue #308 by not to store any records that are not relevant to the daemon.  In this case, it will not store records before browsing the service type. Hence we don't need to clear the cache anyway. Closing this PR for now unless the issue happens again.

PR #342, opened 2025-04-09
feat: extend `DnsAddress` with an `InterfaceId`

### LINKED ISSUE #339: IPv6 link local addresses of a resolved service are not usable without a scope id
I've been experimenting with adding an "Open" button for services like _http._tcp in my mDNS browsing app. While doing so, I ran into a problem when resolving services that only advertise an IPv6 link-local address.

The issue is that if multiple network interfaces are present, the resolved IPv6 link-local address is essentially unusable without a scope ID. Normally, the scope ID can be inferred from the receiving interface, but I'm not entirely confident this can be done reliably in all cases.

For comparison, Avahi includes the scope (as an interface index) when resolving records, which seems necessary to make use of link-local addresses in multi-interface environments.

Would it be possible to include the scope/interface information in the resolved record?

hrzlgnm:
This allows for adding an interface scope to resolved IPv6 link local addresses to resolve #339

### DIFF
--- a/src/dns_parser.rs
+++ b/src/dns_parser.rs
@@ -22,7 +22,7 @@ use std::{
     time::SystemTime,
 };
 
-/// InterfaceId is used to represent the interface indentifer
+/// InterfaceId is used to represent the interface identifier
 #[derive(Clone, Debug, Eq, Hash, PartialEq)]
 pub struct InterfaceId {
     name: String,
--- a/src/dns_parser.rs
+++ b/src/dns_parser.rs
@@ -9,6 +9,8 @@ use crate::log::trace;
 
 use crate::error::{e_fmt, Error, Result};
 
+use if_addrs::Interface;
+
 use std::{
     any::Any,
     cmp,
@@ -19,6 +21,34 @@ use std::{
     str,
     time::SystemTime,
 };
+
+/// InterfaceId is used to represent the interface indentifer
+#[derive(Clone, Debug, Eq, Hash, PartialEq)]
+pub struct InterfaceId {
+    name: String,
+    index: u32,
+}
+
+impl InterfaceId {
+    pub fn index(&self) -> u32 {
+        self.index
+    }
+}
+
+impl fmt::Display for InterfaceId {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{}: {}", self.index, self.name)
+    }
+}
+
+impl From<&Interface> for InterfaceId {
+    fn from(interface: &Interface) -> Self {
+        InterfaceId {
+            name: interface.name.clone(),
+            index: interface.index.unwrap_or_default(),
+        }
+    }
+}
 
 /// DNS resource record types, stored as `u16`. Can do `as u16` when needed.
 ///
@@ -485,12 +515,24 @@ pub trait DnsRecordExt: fmt::Debug {
 pub struct DnsAddress {
     pub(crate) record: DnsRecord,
     address: IpAddr,
+    interface_id: InterfaceId,
 }
 
 impl DnsAddress {
-    pub fn new(name: &str, ty: RRType, class: u16, ttl: u32, address: IpAddr) -> Self {
+    pub fn new(
+        name: &str,
+        ty: RRType,
+        class: u16,
+        ttl: u32,
+        address: IpAddr,
+        interface_id: InterfaceId,
+    ) -> Self {
         let record = DnsRecord::new(name, ty, class, ttl);
-        Self { record, address }
+        Self {
+            record,
+            address,
+            interface_id,
+        }
     }
 
     pub fn address(&self) -> IpAddr {
@@ -520,7 +562,9 @@ impl DnsRecordExt for DnsAddress {
 
     fn matches(&self, other: &dyn DnsRecordExt) -> bool {
         if let Some(other_a) = other.any().downcast_ref::<Self>() {
-            return self.address == other_a.address && self.record.entry == other_a.record.entry;
+            return self.address == other_a.address
+                && self.record.entry == other_a.record.entry
+                && self.interface_id == other_a.interface_id;
         }
         false
     }
@@ -1661,10 +1705,11 @@ pub struct DnsIncoming {
     num_answers: u16,
     num_authorities: u16,
     num_additionals: u16,
+    interface_id: InterfaceId,
 }
 
 impl DnsIncoming {
-    pub fn new(data: Vec<u8>) -> Result<Self> {
+    pub fn new(data: Vec<u8>, interface_id: InterfaceId) -> Result<Self> {
         let mut incoming = Self {
             offset: 0,
             data,
@@ -1678,6 +1723,7 @@ impl DnsIncoming {
             num_answers: 0,
             num_authorities: 0,
             num_additionals: 0,
+            interface_id,
         };
 
         /*
@@ -1945,12 +1991,26 @@ impl DnsIncoming {
                         .boxed(),
                     ),
                     RRType::A => Some(
-                        DnsAddress::new(&name, rr_type, class, ttl, self.read_ipv4()?.into())
-                            .boxed(),
+                        DnsAddress::new(
+                            &name,
+                            rr_type,
+                            class,
+                            ttl,
+                            self.read_ipv4()?.into(),
+                            self.interface_id.clone(),
+                        )
+                        .boxed(),
                     ),
                     RRType::AAAA => Some(
-                        DnsAddress::new(&name, rr_type, class, ttl, self.read_ipv6()?.into())
-                            .boxed(),
+                        DnsAddress::new(
+                            &name,
+                            rr_type,
+                            class,
+                            ttl,
+                            self.read_ipv6()?.into(),
+                            self.interface_id.clone(),
+                        )
+                        .boxed(),
                     ),
                     RRType::NSEC => Some(
                         DnsNSec::new(
--- a/src/service_daemon.rs
+++ b/src/service_daemon.rs
@@ -3115,7 +3115,7 @@ enum Command {
     /// Get the current status of the daemon.
     GetStatus(Sender<DaemonStatus>),
 
-    /// Monitor noticable events in the daemon.
+    /// Monitor noticeable events in the daemon.
     Monitor(Sender<DaemonEvent>),
 
     SetOption(DaemonOption),
--- a/src/service_daemon.rs
+++ b/src/service_daemon.rs
@@ -1618,6 +1618,7 @@ impl Zeroconf {
                     CLASS_IN | CLASS_CACHE_FLUSH,
                     0,
                     address,
+                    intf.into(),
                 ),
                 0,
             );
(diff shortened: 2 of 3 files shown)


src/dns_cache.rs
```diff
@@ -218,8 +218,9 @@ impl DnsCache {
                 let addr = answer_addr.address();
                 if !valid_ip_on_intf(&addr, intf) {
                     debug!(
-                        "add_or_update: answer addr {addr} not in the subnet of {}",
-                        intf.ip()
+                        "add_or_update: answer addr {addr} not in the subnet of {} ({})",
```
hrzlgnm, line 221: I found that addition useful, if it isn't feasible please let me know.

src/service_daemon.rs
```diff
@@ -1344,7 +1344,7 @@ impl Zeroconf {
         if info.is_addr_auto() {
             let selected_addrs = self.selected_addrs(my_ip_interfaces(true));
             for addr in selected_addrs {
-                info.insert_ipaddr(addr);
+                info.insert_scoped_addr(addr, None);
```
hrzlgnm, line 1347: I'm not sure yet if we can eliminate this None and the one in https://github.com/keepsimple1/mdns-sd/pull/342/files#diff-2061a30a0f84a30fe5c20cfb9505c70e8efd1147fc27ebb4deb5e35c53d3a59fR205 which would make the Option superfluous.

src/service_daemon.rs
```diff
@@ -3851,13 +3856,10 @@ mod tests {
         let mut resolved = false;
 
         while let Ok(event) = browse_chan.recv_timeout(timeout) {
-            match event {
```
hrzlgnm, line 3854: Those next 3 changes were suggested by rust clippy for [single_match](https://rust-lang.github.io/rust-clippy/master/index.html#single_match), if not feasible please let me know.

keepsimple1 (commented):
Thanks again for your PR!  I've added a few inline comments. And I also have a high level question:   

Instead of add interface info in each address, should we add the interface info at `ServiceInfo` ?  

For IPv6,  I think it's pretty clear we can / should add the interface at `ServiceInfo` layer because mDNS for IPv6 is in link-local scopes (only).  It is possible that there are two service instances with exact same name but on two different IPv6 interfaces.  Currently we would combine them into one `ServiceInfo` with two addresses, which would be incorrect.  

For IPv4, things are different.  If a host has 2 IPv4 interfaces configured on the same LAN, they will only see one service instance given the same name.  It's ok / making sense to combine into one ServiceInfo.  That said, it's also workable if we return two ServiceInfo records each for one Interface. 

Looking at Avahi, it seems they are reporting service instances based on interfaces. So basically my question is:  shall we do that too?    (Avahi also reports IPv4 and IPv6 separately, but I think we could combine them). 

What do you think?

hrzlgnm:
> Thanks again for your PR! I've added a few inline comments. And I also have a high level question:

Thanks for reviewing, I will address those comments after resolving the high level discussion.

> Instead of add interface info in each address, should we add the interface info at `ServiceInfo` ?
> 
> For IPv6, I think it's pretty clear we can / should add the interface at `ServiceInfo` layer because mDNS for IPv6 is in link-local scopes (only). It is possible that there are two service instances with exact same name but on two different IPv6 interfaces. Currently we would combine them into one `ServiceInfo` with two addresses, which would be incorrect.
> 
> For IPv4, things are different. If a host has 2 IPv4 interfaces configured on the same LAN, they will only see one service instance given the same name. It's ok / making sense to combine into one ServiceInfo. That said, it's also workable if we return two ServiceInfo records each for one Interface.
> 
> Looking at Avahi, it seems they are reporting service instances based on interfaces. So basically my question is: shall we do that too? (Avahi also reports IPv4 and IPv6 separately, but I think we could combine them).
> 
> What do you think?

Yes that makes sense, I thought about this route shorty but didn't yet tried it out. Now it also makes sense to me to separate IPv4 and IPv6 ServiceInfos and per interface like Avahi does. That would make the structure more cleaner. From a library consumer perspective that approach would be more feasible than the route I took initially. The consumer of the library can still `consolidate` those service infos based on the instance name if needed.

keepsimple1:
> Now it also makes sense to me to separate IPv4 and IPv6 ServiceInfos and per interface like Avahi does. That would make the structure more cleaner. 

Could you elaborate more about this separation of IPv4 and IPv6 Serviceinfos ?  My initial thought was that they can always been combined and so this library can do it for the users. But I'd like to understand more from your point of view.

hrzlgnm:
> > Now it also makes sense to me to separate IPv4 and IPv6 ServiceInfos and per interface like Avahi does. That would make the structure more cleaner.
> 
> Could you elaborate more about this separation of IPv4 and IPv6 Serviceinfos ? My initial thought was that they can always been combined and so this library can do it for the users. But I'd like to understand more from your point of view.

Ah, maybe I got something wrong. I envisioned an addition of `InterfaceId` to `ServiceInfo`. Which of course allows for adding both, IPv6 and IPv4 addresses, as long those are received from the same `InterfaceId`. A separation to `IPv4` and `IPv6` for`ServiceInfo` isn't really necessary in that case, I just went to far with making this similar with Avahi.

src/service_info.rs
```diff
@@ -16,6 +16,72 @@ use std::{
...
+
+impl PartialEq<Interface> for InterfaceScope {
+    fn eq(&self, other: &Interface) -> bool {
+        self.name == other.name && self.index == other.index
+    }
+}
+
+/// ScopedAddress is used to represent an address and its interface scope
+#[derive(Clone, Debug, Eq, Hash, PartialEq)]
+pub struct ScopedAddress {
```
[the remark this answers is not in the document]
hrzlgnm: Fair point, but I guess we sort of agreed upon adding InterfaceId to `ServiceInfo` instead.

hrzlgnm:
@keepsimple1: 'm going to add `InterfaceId` to `ServiceInfo` and `ResolvedService` instead of adding more complexity with my first approach using `ScopedAddress` and  `InterfaceScope`.

`InterfaceId` being defined as:

```rust
struct InterfaceId {
      name: String,
      index: Option<u32>,
}
```

If you don't object.

src/dns_parser.rs
```diff
@@ -7,7 +7,10 @@
 #[cfg(feature = "logging")]
 use crate::log::trace;
 
-use crate::error::{e_fmt, Error, Result};
+use crate::{
+    error::{e_fmt, Error, Result},
+    service_info::InterfaceScope,
```
hrzlgnm, line 12: @keepsimple1: `dns_parser` is ought to be a separate module, do we want to provide `InterfaceId` from `dns_parser`. Or would it be better create a separate module `interface_id` which provides `InterfaceId` for `dns_parser` and `service_info`. I'm not that fluent yet in rust best practices about modules, I'm just guessing from experience from other languages and I sort of would prefer the `interface_id` module.

hrzlgnm:
I've started to work on adding an `InterfaceId` to `ServiceInfo` but the following issue arises, if we add that field, we need to initialize the field in `ServiceInfo::new()`.

https://github.com/keepsimple1/mdns-sd/blob/4192b1a9bdfa8cc003e2971b3a0abead481e2deb/src/service_info.rs#L94

But how do we want to do that without introducing a breaking change?

hrzlgnm:
> I've started to work on adding an `InterfaceId` to `ServiceInfo` but the following issue arises, if we add that field, we need to initialize the field in `ServiceInfo::new()`.
> 
> https://github.com/keepsimple1/mdns-sd/blob/4192b1a9bdfa8cc003e2971b3a0abead481e2deb/src/service_info.rs#L94
> 
> But how do we want to do that without introducing a breaking change?

I see the following option here:
* add  `with_interface_id()` constructor which we then would use when resolving, so probably making it also `pub(crate)`. Which also makes the interface_id an Option<InterfaceId>.

src/dns_parser.rs
keepsimple1: I think it's okay to define `InterfaceId` in `dns_parser`.  We can refactor the module later once we have clearer understanding.

keepsimple1:
> > But how do we want to do that without introducing a breaking change?
> 
> I see the following option here:
> 
> * add  `with_interface_id()` constructor which we then would use when resolving, so probably making it also `pub(crate)`. Which also makes the interface_id an `Option<InterfaceId>`.

I'm thinking maybe this is a good time for us to move towards using `ResolvedService` when resolving? Such option could be:  

- define a new variant in `ServiceEvent`, something like `ServiceEvent::ServiceResolvedIntf(ResolvedService)`
- add `interface_id` into `ResolvedService`

When a service is resolved, we will send both existing `ServiceResolved(ServiceInfo)` and the new `ServiceResolvedIntf(ResolvedService)` to the querier.  If the new variant works well,  we can deprecate the `ServiceResolved(ServiceInfo)` over time, and let `ServiceInfo` only used by the register / publisher.

Of course, this still a breaking change because `ServiceEvent` is not marked as `#[non_exhaustive]` yet.  But I hope its impact is relatively small. 

Thoughts?

keepsimple1:
> `InterfaceId` being defined as:
> 
> ```rust
> struct InterfaceId {
>       name: String,
>       index: Option<u32>,
> }
> ```
> 
> If you don't object.

No objections. The only thing I was curious is that: should `index` be `Option`?  I understand that when `name` is invalid, the OS will not return an index in `if-addrs` crate.  But if `name` is valid, shouldn't we always have `index`?

keepsimple1:
I think this PR might be too involved. And since the full diff will introduce some breaking changes, I am wondering if we can split it into two steps:  
1) merge in internal changes first, namely only `DnsAddress` and `InterfaceId` changes,  and later  
2) decide the public API approach and make related changes. 

Would you like to use this PR to commit the 1) stuff first?  Meantime we can think more about 2) and also do some experiments.

hrzlgnm:
> I think this PR might be too involved. And since the full diff will introduce some breaking changes, I am wondering if we can split it into two steps:
> 
>     1. merge in internal changes first, namely only `DnsAddress` and `InterfaceId` changes,  and later
> 
>     2. decide the public API approach and make related changes.
> 
> 
> Would you like to use this PR to commit the 1) stuff first? Meantime we can think more about 2) and also do some experiments.

Yes sounds reasonable, I'll only touch `DnsAddress` and `InterfaceId` with this PR.

src/service_daemon.rs
hrzlgnm: I've reverted all unrelated changes, to avoid adding too much refactoring to this PR

src/dns_parser.rs
```diff
@@ -20,6 +22,34 @@ use std::{
     time::SystemTime,
 };
 
+/// InterfaceId is used to represent the interface indentifer
```
keepsimple1, line 25: typo: "indentifer" -> "identifier"

keepsimple1 (commented):
LGTM! Only one minor comment about a typo. Thanks!

src/dns_parser.rs
hrzlgnm: Fixed with https://github.com/keepsimple1/mdns-sd/pull/342/commits/35116be8e22fda1dbae5d9a9514f25179c6ce3f0, thanks for reviewing.

hrzlgnm: I also fixed some more minor spelling issues with `typos-cli` in https://github.com/keepsimple1/mdns-sd/pull/342/commits/ba742edaf0105cdb728ca5da9f63b4d2a3e3d71a

keepsimple1 (approved):
LGTM! Thanks for updates!

Issue #351, opened 2025-04-25
Properties disappearing for long running discovery ServiceDaemons


hgomersall:
I'm not completely sure that the issue title describes the problem accurately and could probably describe my erroneous usage of the library, but in any case, it would be good to clarify the situation.

If we have a simple mdns-sd server, that adds some properties to the ServiceInfo, it seems that if one tries to reuse a ServiceDaemon on the client side (e.g. create one once and hold it in a static `OnceLock`), after a bit of time, the service seems to return as expected but there are no properties attached to the service (i.e. `get_properties()` returns empty).

Following is a minimal case that shows the usage. If you replace the OnceLock ServiceDaemon with a new ServiceDaemon on every call, it works as expected (also using the shutdown code at the end).
```rust
use mdns_sd::{IfKind, ServiceDaemon, ServiceEvent};
use std::net::IpAddr;
use std::sync::OnceLock;
use std::time::Duration;

const SERVICE_TYPE: &str = "_My_Module._tcp.local.";

static SERVICE_DAEMON: OnceLock<ServiceDaemon> = OnceLock::new();

#[derive(Debug, Clone, PartialEq, Eq)]
pub struct MyModule {
    pub addr: IpAddr,
}


pub(crate) fn discover_modules(
    wait_time: std::time::Duration,
) {
    // Create a daemon
    let mdns = SERVICE_DAEMON
        .get_or_init(|| ServiceDaemon::new().expect("Could not create service daemon"));

    //let mdns = ServiceDaemon::new().unwrap();

    // We currently disable IPv6.
    mdns.disable_interface(IfKind::IPv6).unwrap();

    let receiver = mdns.browse(SERVICE_TYPE).expect("Failed to browse");

    let now = std::time::Instant::now();

    let deadline = now + wait_time;

    while let Ok(event) = receiver.recv_deadline(deadline) {
        match event {
            ServiceEvent::ServiceResolved(info) => {
                let properties = info.get_properties();
                dbg!(properties);

                for addr in info.get_addresses() {

                    let new_module = MyModule {
                        addr: *addr,
                    };
                    dbg!(new_module);

                }
            }
            _ => {
            }
        }
    }

    //let shutdown_channel = mdns.shutdown().unwrap();
    //let shutdown_state = shutdown_channel
    //    .recv_timeout(Duration::from_secs(1))
    //    .unwrap();

    //dbg!(shutdown_state);
}
fn main() {

[...]

PR #355, opened 2025-04-28
bugfix: TXT records should use OTHER_TTL same as PTR

keepsimple1:
This is found when debugging issue #351 .  `TXT` records are not refreshed by the client because the server sets TTL to be 120 seconds, which should be 4500 seconds as PTR records. 

I manually tested this diff and saw the issue #351 is gone. Will work on a test case. 

Update: added a test case and some refactoring.

### DIFF
--- a/src/dns_parser.rs
+++ b/src/dns_parser.rs
@@ -1494,6 +1494,11 @@ impl DnsOutgoing {
         &self.questions
     }
 
+    /// For testing purposes only.
+    pub(crate) fn _answers(&self) -> &[(DnsRecordBox, u64)] {
+        &self.answers
+    }
+
     pub fn answers_count(&self) -> usize {
         self.answers.len()
     }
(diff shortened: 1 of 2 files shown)


hgomersall:
To be clear, extending this to 4500 seconds should result in a correct refresh because, presumably, there is a refresh process inside that time window? That is, it should now live arbitrarily long if the service is still up?

keepsimple1:
> To be clear, extending this to 4500 seconds should result in a correct refresh because, presumably, there is a refresh process inside that time window? 

yes, but it's because PTR for the service has the same TTL and will be [refreshed](https://github.com/keepsimple1/mdns-sd/blob/07750cb2bda6918d2ab93cd92017ae18ee4abb14/src/service_daemon.rs#L2967). The response of PTR query (should) includes TXT records as well. (Another PR #354 will refresh TXT records specifically)

> That is, it should now live arbitrarily long if the service is still up?

Yes.

PR #361, opened 2025-06-06
bugfix: set multicast TTL to 255

### LINKED ISSUE #360: Should the mDNS packet TTL be 255?
In Section 11 of the mDNS RFC 6762:

>  All Multicast DNS responses (including responses sent via unicast)
>    SHOULD be sent with IP TTL set to 255.  This is recommended to
>    provide backwards-compatibility with older Multicast DNS queriers
>    (implementing a draft version of this document, posted in February
>    2004) that check the IP TTL on reception to determine whether the
>    packet originated on the local link.  These older queriers discard
>    all packets with TTLs other than 255.

I noticed that the socket created by mdns-sd never explicitly sets the TTL: https://github.com/keepsimple1/mdns-sd/blob/main/src/service_daemon.rs#L684

If this is not by choice, would you accept a PR that changes behavior to match the RFC? I'll add fallible calls to `set_multicast_ttl_v4(255)` and `set_multicast_hops_v6(255)` on the sockets.

sameer:
Fixes #360

I'm not sure this should be categorized as a bugfix, but it is a behavioral change worth calling out.

### DIFF
--- a/src/service_daemon.rs
+++ b/src/service_daemon.rs
@@ -698,6 +698,9 @@ fn new_socket_bind(intf: &Interface, should_loop: bool) -> Result<MioUdpSocket> 
             sock.set_multicast_if_v4(ip)
                 .map_err(|e| e_fmt!("set multicast_if on addr {}: {}", ip, e))?;
 
+            sock.set_multicast_ttl_v4(255)
+                .map_err(|e| e_fmt!("set set_multicast_ttl_v4 on addr {}: {}", ip, e))?;
+
             if !should_loop {
                 sock.set_multicast_loop_v4(false)
                     .map_err(|e| e_fmt!("failed to set multicast loop v4 for {ip}: {e}"))?;
@@ -723,6 +726,9 @@ fn new_socket_bind(intf: &Interface, should_loop: bool) -> Result<MioUdpSocket> 
             // Set IPV6_MULTICAST_IF to send packets.
             sock.set_multicast_if_v6(intf.index.unwrap_or(0))
                 .map_err(|e| e_fmt!("set multicast_if on addr {}: {}", ip, e))?;
+
+            sock.set_multicast_hops_v6(255)
+                .map_err(|e| e_fmt!("set set_multicast_hops_v6 on addr {}: {}", ip, e))?;
 
             // We are not sending multicast packets to test this socket as there might
             // be many IPv6 interfaces on a host and could cause such send error:


src/service_daemon.rs
```diff
@@ -698,6 +698,9 @@ fn new_socket_bind(intf: &Interface, should_loop: bool) -> Result<MioUdpSocket>
             sock.set_multicast_if_v4(ip)
                 .map_err(|e| e_fmt!("set multicast_if on addr {}: {}", ip, e))?;
 
+            sock.set_multicast_ttl_v4(255)
+                .map_err(|e| e_fmt!("set set_multicast_ttl_v4 on addr {}: {}", ip, e))?;
+
```
keepsimple1, line 703: There is one thing I didn't realize earlier: the section 11 in RFC 6762 only mentions "Multicast DNS responses".  What about queries?  

The TTL set on the socket would apply for all packets going out, including responses and queries. I'm wondering if TTL 255 has any side effects for the queries (its default TTL is 1).

keepsimple1 (commented):
Thanks for your PR! Please see comment inline.

src/service_daemon.rs
keepsimple1: I think we can set the IP TTL in this function instead: 
https://github.com/keepsimple1/mdns-sd/blob/22a2688d6a15c7d0d5c535c7f1c76e891abd3de1/src/service_daemon.rs#L3258

where we can check `qtype`. If it's a response, set IP TTL to 255, otherwise set it to 1. 

And since this is a `Should` in the RFC, I think we can handle set TTL error gracefully, i.e. log a debug message but not to fail `send_dns_outgoing`.

sameer: Using Wireshark, I confirmed that all mDNS traffic on a local network is supposed to set the TTL/hop limit to 255. Wireshark even notes that
 the TTL set by my router is incorrectly `1` and not `255` like the RFC says.

Screenshot 2025-06-16 at 18 57 04

sameer: Setting the multicast TTL/hops is the most important part. To keep this PR simple, I won't change unicast TTL.

The way it's currently being set in this PR looks right to me:

- Affects all outbound multicast
- Wireshark notes that the TTL for a multicast query should also be 255

keepsimple1: I am okay with the the change although RFC 6762 does not specify TTL for query packets.  I don't any obvious problems per my knowledge.

Regarding unicast, note that RFC does mention it as well in [section 11](https://datatracker.ietf.org/doc/html/rfc6762#section-11): 
```
All Multicast DNS responses (including responses sent via unicast)
   SHOULD be sent with IP TTL set to 255.   
```

Currently we are good with multicast only as we don't support unicast yet.

keepsimple1 (approved):
All things considered, I think it looks good.

Issue #358, opened 2025-05-19
IPv6 link-local addresses


parasyte:
The `ServiceInfo` struct returned in `ServiceEvent::ServiceResolved` can include link-local IPv6 addresses without any scope/zone index information, making them useless. IPv6 requires link-local addresses, so you are pretty much guaranteed to receive these when the service browser daemon is active.

In #242, the daemon itself was fixed to allow link-local addresses to be reachable by the daemon. The problem here is that, as a client, I need the same information to be able to create a valid socket with these link-local addresses. The behavior and some background details were mentioned in this comment when IPv6 support was initially being added: https://github.com/keepsimple1/mdns-sd/pull/130#issuecomment-1750311788

My current workaround is ignoring IPv6 link-local addresses. But I would be able to use them if the interface number that the addresses were discovered on was included in the event.

keepsimple1:
A similar issue was discussed recently in #342 , and we got a basic (internal) struct committed to include the interface number, while left the full solution not decided yet. 

It seems now it's a good time to come back and work out a full solution ;-). Early one of my concerns was: only a client (i.e. querier) would use the interface number inside an address, and that would conflict with the usage of `ServiceInfo` on the server side.  But now I think probably even the server side can benefit from specifying interface number / scope id for publishing its service on IPv6.

And it's likely be a breaking change in API, but I think we can make it as minimum as possible.

parasyte:
Thanks! I hadn't seen #342 yet. I'm going over it, now.

BTW, here are a few related conversations:

- https://github.com/avahi/avahi/issues/110
- https://unix.stackexchange.com/questions/583794/how-to-create-peer-to-peer-ipv6-network-using-link-local-addresses

I'm indifferent to any changes that might be necessary for this. I know, for instance, that the `ServiceInfo` is used in two places: 1. to register the service and 2. to receive service update events from the daemon. It absolutely makes sense to include the interface index in the latter. Splitting these into different structs would be acceptable to me, but I understand if others feel differently.

Additionally, I don't know how useful it is to include the interface index in the service registration. The interface information is local to the host that is registering the service, and it cannot be used by any remote peers.

keepsimple1:
> Additionally, I don't know how useful it is to include the interface index in the service registration. The interface information is local to the host that is registering the service, and it cannot be used by any remote peers.

[...]

PR #362, opened 2025-06-09
feat: new API to use ResolvedService instead of ServiceInfo for resolved service event

keepsimple1:
This is to address issue #358.  The basic approach is: 

- Keep it backward compatible for `ServiceEvent::ServiceResolved` event, which returns `ServiceInfo`.
- Introduce a new event `ServiceEvent::ServiceDetailed` to return `ResolvedService`.  Update `ResolvedService.addresses` to be `HashSet<HostIp>`, where `HostIp` is a wrapper of `IpAddr` that includes `scope_id` for IPv6 addresses.

To opt in to receive `ServiceEvent::ServiceDetailed` instead of `ServiceEvent::ServiceResolved`, the user would call `ServiceDaemon::use_service_detailed(true)` of its client daemon.

So, in short, to use IPv6 address with scope (interface id), one needs to use the new `ServiceDetailed` event and `ResolvedService` struct instead of `ServiceInfo`.   In the long term, I hope to deprecate the uses of `ServiceInfo` on the client (querier) side.

Also note, `HostnameResolutionEvent` is also updated to use `HostIp` instead of `IpAddr`.  I think it's better to make this breaking change at the same time.

### DIFF
--- a/src/dns_parser.rs
+++ b/src/dns_parser.rs
@@ -47,6 +47,83 @@ impl From<&Interface> for InterfaceId {
     }
 }
 
+/// An IPv4 address used in `HostIp`.
+#[derive(Debug, Clone, Eq, PartialEq, Hash)]
+pub struct HostIpV4 {
+    addr: Ipv4Addr,
+}
+
+/// An IPv6 address with scope_id (interface identifier).
+#[derive(Debug, Clone, Eq, PartialEq, Hash)]
+pub struct HostIpV6 {
+    addr: Ipv6Addr,
+    scope_id: InterfaceId,
+}
+
+/// A host IP address, either IPv4 or IPv6, that supports scope_id for IPv6.
+#[derive(Debug, Clone, Eq, PartialEq, Hash)]
+#[non_exhaustive]
+pub enum HostIp {
+    V4(HostIpV4),
+    V6(HostIpV6),
+}
+
+impl HostIp {
+    pub const fn to_ip_addr(&self) -> IpAddr {
+        match self {
+            HostIp::V4(v4) => IpAddr::V4(v4.addr),
+            HostIp::V6(v6) => IpAddr::V6(v6.addr),
+        }
+    }
+
+    pub const fn is_ipv4(&self) -> bool {
+        matches!(self, HostIp::V4(_))
+    }
+
+    pub const fn is_ipv6(&self) -> bool {
+        matches!(self, HostIp::V6(_))
+    }
+}
+
+impl From<IpAddr> for HostIp {
+    fn from(ip: IpAddr) -> Self {
+        match ip {
+            IpAddr::V4(v4) => HostIp::V4(HostIpV4 { addr: v4 }),
+            IpAddr::V6(v6) => HostIp::V6(HostIpV6 {
+                addr: v6,
+                scope_id: InterfaceId::default(),
+            }),
+        }
+    }
+}
+
+impl From<&Interface> for HostIp {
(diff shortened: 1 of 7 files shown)


hrzlgnm:
I guess we also should consider updating the query example to use the new proposed API.

keepsimple1:
> I guess we also should consider updating the query example to use the new proposed API.

I've updated the query example.

hrzlgnm:
Can we make the struct `dns_parser::InterfaceId`  `pub` in lib.rs? Otherwise consumers cannot implement a `From` trait for that struct if needed?

keepsimple1:
> Can we make the struct `dns_parser::InterfaceId` `pub` in lib.rs? Otherwise consumers cannot implement a `From` trait for that struct if needed?

Exported `InterfaceId`. I might missed something but I don't think Rust allows external consumers implement `From` trait for `InterfaceId` per [the orphan rule](https://doc.rust-lang.org/book/ch10-02-traits.html#implementing-a-trait-on-a-type).

hrzlgnm:
> > Can we make the struct `dns_parser::InterfaceId` `pub` in lib.rs? Otherwise consumers cannot implement a `From` trait for that struct if needed?
> 
> Exported `InterfaceId`. I might missed something but I don't think Rust allows external consumers implement `From` trait for `InterfaceId` per [the orphan rule](https://doc.rust-lang.org/book/ch10-02-traits.html#implementing-a-trait-on-a-type).

Thanks, I was maybe a bit imprecise on which direction the trait goes. I wanted to implement a `From<InterfaceId>` for another struct in my crate which should be possible now.

src/dns_parser.rs
```diff
@@ -50,6 +47,83 @@ impl From<&Interface> for InterfaceId {
...
+    }
+}
+
+impl fmt::Display for HostIp {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            HostIp::V4(v4) => write!(f, "{}", v4.addr),
+            HostIp::V6(v6) => {
+                if v6.scope_id.index != 0 {
+                    write!(f, "{}%{}", v6.addr, v6.scope_id.name)
```
hrzlgnm, line 118: I'm not sure whether address%index would be more feasible, as for example [Windows](https://learn.microsoft.com/en-us/windows/win32/winsock/link-local-and-site-local-addresses-2) does not use interface names but rather an interface index as a scope id.

keepsimple1:
@oysteintveit-nordicsemi may I ask you help review the changes to `HostnameResolutionEvent` where `AddressesFound` and `AddressesRemoved` changes from using `IpAddr` to using `HostIp` ?   Is it a positive change from `HostnameResolutionEvent` point of view?

This is a breaking change for `HostnameResolutionEvent`. Are there some methods that should be implemented for `HostIp` so that your code change is minimum?

h7x4:
Hello, this is @oysteintveit-nordicsemi from my non-work account, thank you for reaching out! I no longer work there so I don't have access to the code, but I can have a look at it and see if I can tell from memory.

I currently don't have access to a computer either (not until next week), but I had a quick skim through the diff and if the only thing that changed is the wrapping struct and the address inside is still accessible to the library user in some way, I think it'll be fine 😁

keepsimple1:
> I currently don't have access to a computer either (not until next week), but I had a quick skim through the diff and if the only thing that changed is the wrapping struct and the address inside is still accessible to the library user in some way, I think it'll be fine 😁

@h7x4 thank you! I appreciated that.

PR #377, opened 2025-07-31
feat: Remove unneeded multicast send tracking

hrzlgnm:
This became obsolete with #373, as we now only use at max one IPv4 and at max one IPv6 address per interface.

### DIFF
--- a/src/service_daemon.rs
+++ b/src/service_daemon.rs
@@ -44,7 +44,7 @@ use crate::{
     Receiver, ResolvedService, TxtProperties,
 };
 use flume::{bounded, Sender, TrySendError};
-use if_addrs::{IfAddr, Interface};
+use if_addrs::Interface;
 use mio::{event::Source, net::UdpSocket as MioUdpSocket, Interest, Poll, Registry, Token};
 use socket_pktinfo::PktInfoUdpSocket;
 use std::{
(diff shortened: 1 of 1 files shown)


hrzlgnm:
I've played around with an added Counter for skipped multicast, and additional added addresses on the network interface and the counter  never actually increased so I guess it is safe to remove that feature.

src/service_daemon.rs
```diff
@@ -1578,18 +1547,9 @@ impl Zeroconf {
     /// Returns the list of interface IPs that sent out the announcement.
     fn send_unsolicited_response(&mut self, info: &mut ServiceInfo) -> Vec<IpAddr> {
         let mut outgoing_addrs = Vec::new();
-        // Send the announcement on one interface per ip version.
-        let mut multicast_sent_trackers = HashSet::new();
-
         let mut outgoing_intfs = Vec::new();
 
         for intf in self.my_intfs.iter() {
```
keepsimple1, line 1552: Although I would love to see any removal of unneeded code, I don't find the original issue (in #239) was gone. 

It's true we no longer have separate `sock` for each intf, and use the common `ipv4_sock` and `ipv6_sock`.  But we are still looping through `self.my_intfs` here, and if an interface has multiple addresses, multiple `intf` will show up and we will send out packet for each of them. 

Maybe I've missed something, please let me know.

hrzlgnm: As far as I understand the following loop: https://github.com/keepsimple1/mdns-sd/blob/8cb1a59387d5a3baaa6087d09b386be8783824eb/src/service_daemon.rs#L969, we only add one IPv4 and one IPv6 address per interface, as a second bind to the respective address family fails and is skipped. That's what I've observed under Linux, not sure about the behavior of other platforms.

keepsimple1: You are correct 👍 . I didn't realize that case, which would render the concerned code unneeded.

keepsimple1 (approved):
LGTM, thanks!

PR #380, opened 2025-08-06
refactoring: define our own interface struct to allow multiple addresses

keepsimple1:
This is triggered by #377 as currently we would skip 2nd IP address on an interface due to `join_multicast_group` error on the same interface.   Strictly speaking, this is wrong and it can cause confusion in debugging as `addr 1` is in `ZeroConf.my_intfs` but `addr 2` (on the same interface) is not. 

Also, currently we directly use `Interface` from `if-addrs` crate, but it is per address, not truly per interface. I took this opportunity to define a new interface struct for us `MyIntf` that represents a network interface that supports multiple addresses.

The diff is very extensive but conceptually really just `Interface` is replaced by `MyIntf`.

### DIFF
--- a/src/dns_cache.rs
+++ b/src/dns_cache.rs
@@ -6,10 +6,9 @@
 use crate::log::{debug, trace};
 use crate::{
     dns_parser::{DnsAddress, DnsPointer, DnsRecordBox, DnsSrv, InterfaceId, RRType},
-    service_info::{split_sub_domain, valid_ip_on_intf, valid_two_addrs_on_intf},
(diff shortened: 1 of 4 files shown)


keepsimple1:
@hrzlgnm I would like invite you to take a look at this, and possibly run some quick test with this PR if possible. It's OK to skip if no time.

hrzlgnm:
@keepsimple1 Did some quick tests on Linux and Windows. LGTM.

keepsimple1:
Thank you @hrzlgnm for reviewing!

PR #388, opened 2025-08-29
feat: support cache only browsing

twizansk:


keepsimple1 (commented):
Thanks for your PR.  Please add a reference to the original issue number in Description. Comments inline.

src/service_daemon.rs
```diff
@@ -493,6 +493,17 @@ impl ServiceDaemon {
         )))
     }
 
+    /// If `accept` is true, accept and cache all responses. If `accept` is false, accept only responses
+    /// matching queries that we have initiated.
+    ///
```
keepsimple1, line 498: Could you also add something like "The default is false" in the doc comments for this fn?

twizansk: done

src/service_daemon.rs
```diff
@@ -303,11 +303,11 @@ impl ServiceDaemon {
     ///
     /// When a new instance is found, the daemon automatically tries to resolve, i.e.
     /// finding more details, i.e. SRV records and TXT records.
-    pub fn browse(&self, service_type: &str) -> Result<Receiver<ServiceEvent>> {
+    pub fn browse(&self, service_type: &str, cache_only: bool) -> Result<Receiver<ServiceEvent>> {
```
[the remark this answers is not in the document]
twizansk: done

twizansk:
> Thanks for your PR. Please add a reference to the original issue number in Description. Comments inline.

updated the pr description

src/service_daemon.rs
```diff
@@ -4749,4 +4799,58 @@ mod tests {
...
+        let new_ttl = 3; // for testing only.
+        my_service._set_other_ttl(new_ttl);
+
+        // register my service
+        let mdns_server = ServiceDaemon::new().expect("Failed to create mdns server");
+        let result = mdns_server.register(my_service);
+        assert!(result.is_ok());
+
+        let mdns_client = ServiceDaemon::new().expect("Failed to create mdns client");
+        let browse_chan = mdns_client.browse_cache(service_type).unwrap();
```
keepsimple1, line 4834: I am wondering, why this `browse_cache` succeeds even when `accept_unsolicited` was not called for `mdns_client` ?   

By default, PTR packets would be dropped as `is_for_us` would be false before calling `browse_cache`.

keepsimple1 (commented):
Thanks for your updates! Overall looks good to me. Only one curious question inline.

src/service_daemon.rs
keepsimple1: > if browse_cache was called first and then the unsolicited announcement is received then the next browse_cache succeeds. Is that the case you're referring to?

Yes.  Btw I tried out your diff locally and confirmed it indeed was what happened.  I think it was because the server took around 1 second for probing,  so that `browse_cache` command was handled before the first announcement packet received by the client. 

To remove that timing factor, we can add a small `sleep()` (2 seconds) before `browse_cache` to let the service announcement happens before `browse_cache`.  I tried locally and indeed the test fails.

So my suggestion for the test:

1. Add `std::thread::sleep(Duration::from_secs(2));` before calling `browse_cache`.
2. Call `accept_unsolicited(true).unwrap()` after creating the client so that the announcement will be cached and the test would pass.

twizansk: done.  also, added a second test for the "browse_cache first" case

keepsimple1 (approved):
Thank you for the updates, LGTM!

PR #398, opened 2025-10-27
feat: service registration with granular iface/ip conditions

twizansk:
suggested update to mdns-sd to support granular service registration.  

We would find it very useful to be able to register services with specific conditions:
- associate them with with specific interfaces.
- return only link-local ips

@keepsimple1 if this looks like a reasonable update to you, I can finish up the pr (cleanup, tests, etc.) and take it out of draft

issue: https://github.com/keepsimple1/mdns-sd/issues/399

### DIFF
--- a/tests/mdns_test.rs
+++ b/tests/mdns_test.rs
@@ -690,11 +690,9 @@ fn service_with_ipv4_only() {
 }
 
 #[test]
-fn service_with_ipv6_intf() {
+fn service_ipv6_link_local_only() {
     // Create a daemon
     let d = ServiceDaemon::new().expect("Failed to create daemon");
-
-    // Register a service with a name len > 15.
     let service_ipv6_intf = "_test_ipv6_intf._udp.local.";
     let host_name = "my_host_ipv6_intf.local.";
     let host_ipv4 = "";
@@ -714,7 +712,7 @@ fn service_with_ipv6_intf() {
     let result = d.register(my_service);
     assert!(result.is_ok());
 
-    // Browse for a service and verify all addresses are IPv4.
+    // Browse for a services. Verify all addresses are link-local IPv6.
     let browse_chan = d.browse(service_ipv6_intf).unwrap();
     let timeout = Duration::from_secs(2);
     let mut resolved = false;


keepsimple1:
Thanks for your PR! At the high level I'm open to add such granular conditions. Please feel free to move it out of draft. I've a couple of comments inline.

twizansk:
pr is out of draft

src/service_info.rs
```diff
@@ -201,6 +210,21 @@ impl ServiceInfo {
         self.requires_probe = enable;
     }
 
+    /// Set whether the service is restricted to link-local addresses.
+    ///
+    /// By default, it is false.
+    pub fn set_is_link_local_only(&mut self, is_link_local_only: bool) {
```
keepsimple1: I think `set_link_local_only` is clear enough and shorter. Can we use that?

src/service_info.rs
```diff
@@ -201,6 +210,21 @@ impl ServiceInfo {
...
+        self.is_link_local_only = is_link_local_only;
+    }
+
+    /// Set the supported interfaces for this service.
+    ///
+    /// Only ips associated with the provided interfaces will be published.  If set to None (default),
+    /// all interfaces are supported
+    pub fn set_supported_intfs(&mut self, intfs: Option<Vec<String>>) {
```
keepsimple1: I think `set_interfaces` (or `set_intfs`) is clear enough and shorter. Can we use that?

src/service_info.rs
```diff
@@ -422,6 +446,28 @@ impl ServiceInfo {
             txt_properties: self.txt_properties,
         }
     }
+
+    fn is_address_supported(&self, addr: &IpAddr, my_intf: &MyIntf) -> bool {
+        let is_valid_ip_on_intf = my_intf.addrs.iter().any(|x| valid_ip_on_intf(addr, x));
```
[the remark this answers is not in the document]
twizansk: unfortunately ipv6 link local addresses on a specific iface is exactly the use case we care about so I don't think my current solution will serve us :(

Some ideas:

**option 1:  retain the scope information everywhere**

- update ```ServiceInfo.addresses``` to hold a new interface (call it ```MaybeScopedIP```). basically a generalization of  ```ScopedIP```where the scope is optional.  Or maybe it's just an enum type
- in ```ServiceInfo::new``` everything gets parsed to ```MaybeScopedIP``` instead of ```IPAddr```.  clients can provide ```ScopedIP``` s directly or even strings of the form ```fe80::260:8ff:fe52:f9d8%<iface>```
- backward compatibility:  clients can still provide ```IPAddr``` or unscoped strings.  any downstream code needs to handle the case where no scope is provided
- new clients can choose to provide scoped addresses instead of ```IpAddr```.  any downstream code should handle those correctly.  interface filters will work as expected.

**option 2: filter the interfaces on insert**

- update ```ServiceInfo.insert_ipaddr``` to accept the iface along with the address.  both places that call ```insert_ipaddr``` have the interface information available: [herre](https://github.com/keepsimple1/mdns-sd/blob/6c01cf65a399993e231cb8b5410fb8aa64864339/src/service_daemon.rs#L1659) and [here](https://github.com/keepsimple1/mdns-sd/blob/6c01cf65a399993e231cb8b5410fb8aa64864339/src/service_daemon.rs#L1715) so this should be straightforward, I think
- ```ServiceInfo.insert_ipaddr``` inserts only addresses on supported ifaces
- This does not deal with the case where addresses were provided explicitly but at least covers the ```auto_addr``` case

I think option 1 moves us toward the "correct" treatment of ipv6 and even provides a path to deprecating the old behavior.  option 2 is probably less disruptive but does not cover some important cases

I'm happy to work on either (or some third option I haven't thought of).  thoughts?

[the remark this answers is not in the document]
twizansk: >maybe we could filter the interfaces in [annouce_service_on_intf](https://github.com/keepsimple1/mdns-sd/blob/8071f3d5a4c5b9513910cf2c1951099d8210d0c1/src/service_daemon.rs#L4060) such as if info has selected interfaces, then only announce on those interfaces

I think that doesn't actually solve the problem?

assume a host has the following ips:
```fe80::1``` on eth0
```fe80::2``` on wlan0

and we set the supported interfaces for the service:
```
info.set_interfaces(&["eth0"]);
```

when ```announce_service_on_intf``` is called with the interface "eth0",  ```get_addrs_on_my_intf_v6``` will incorrectly return ```fe80::2``` along with ```fe80::1```.  since eth0 is supported, both ips will then be announced on that iface.

I'll put together an implementation of option 2 and I think I can satisfy all the requirements you listed above.

File: src/service_info.rs (hunk 2 above)
Comment (twizansk) [reply src/service_info.rs]: updated

File: src/service_info.rs [hunk 4]
```diff
@@ -201,6 +210,21 @@ impl ServiceInfo {
...
+    /// Set whether the service is restricted to link-local addresses.
+    ///
+    /// By default, it is false.
+    pub fn set_link_local_only(&mut self, is_link_local_only: bool) {
+        self.is_link_local_only = is_link_local_only;
+    }
+
+    /// Set the supported interfaces for this service.
+    ///
+    /// Only ips associated with the provided interfaces will be published.  If set to None (default),
```
keepsimple1: In my understanding, I'd define the semantics of this function as: 

"The service will only be published on the provided interfaces. "

(i.e.  In essence, it's about the interfaces, not about the addresses.  Even though, if it's enable_auto_addr,  the effect is only IP addrs associated with the interfaces are published)

src/service_info.rs
```diff
@@ -201,6 +210,21 @@ impl ServiceInfo {
...
+    /// all interfaces are supported
+    pub fn set_interfaces(&mut self, intfs: Option<Vec<String>>) {
```
keepsimple1: I'm wondering if we should use `Vec<IfKind>` for `intfs` parameter.  (`IfKind` is an existing public type). It's more flexible, can remove the need of `Option`, but also more complex.  Your thoughts?

src/service_info.rs
```diff
@@ -79,6 +80,12 @@ pub struct ServiceInfo {
 
     /// Whether we need to probe names before announcing this service.
     requires_probe: bool,
+
+    /// If set, the service is only exposed on these interfaces
+    supported_intfs: Box<Option<Vec<String>>>,
```
keepsimple1: No need to use `Box` here, since `Vec` already handles heap allocation for its data. 

Also, I wonder if we could use the type `Vec<IfKind>`

keepsimple1 (commented):
If possible (and ready), could you also add integration test for the new public APIs?

src/service_info.rs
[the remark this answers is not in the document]
keepsimple1: If your implementation supports your use case already, no need to worry about `add_addr_with_if_name`. ;-) I will look into that later.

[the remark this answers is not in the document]
twizansk: thanks.  I'm also happy to keep contributing.  this is a great library and already surpasses avahi for our needs :)

src/service_info.rs
twizansk: unfortunately the new ```ServiceInfo``` fields put the ```Command``` enum over the 200 byte threshold of clippy's [large_enum_variant](https://rust-lang.github.io/rust-clippy/master/index.html#large_enum_variant) warning.  But agreed, a Box is probably overkill.  Instead I disabled the warning for that enum.

src/service_info.rs
twizansk: updated

twizansk:
> If possible (and ready), could you also add integration test for the new public APIs?

added.  FYI I noticed that the tests don't run on the ubuntu build.  the workflow file has
```
- name: Run tests with debugs
   if: matrix.os == 'ubuntu-20.04' || matrix.os == 'macos-14'
```
but we're running on ubuntu 22.04

keepsimple1:
> but we're running on ubuntu 22.04

You can update the workflow to use ubuntu 22.04.

src/service_info.rs
keepsimple1: > another alternative is to box the whole ServiceInfo [here](https://github.com/keepsimple1/mdns-sd/pull/398/files#diff-210d5927cee6901a10f55f2b88f90e6a2348a68021b2301675f010588a98c043L3494). thinking about it, that actually seems like the cleanest option. let me know what you prefer

Yes, I prefer this approach. Thanks!

twizansk:
> > but we're running on ubuntu 22.04
> 
> You can update the workflow to use ubuntu 22.04.

getting test failures on 22.04: test_disable_interface_cache.  tested on mainline and still failing (admittedly I tested on 24.04 not 22.04) so it seems unrelated. ok if we leave this as a follow-up?

twizansk:
> > > but we're running on ubuntu 22.04
> > 
> > 
> > You can update the workflow to use ubuntu 22.04.
> 
> getting test failures on 22.04: test_disable_interface_cache. tested on mainline and still failing (admittedly I tested on 24.04 not 22.04) so it seems unrelated. ok if we leave this as a follow-up?

FWIW, on my setup looks like it's failing because, in the test, we disabled [only the ipv4 interface address](https://github.com/keepsimple1/mdns-sd/blob/7293e7c5d1bbb7937adabb341a0a388c397b2779/tests/mdns_test.rs#L701)

however, the iface returns an ipv6 address as well which does not get disabled.

[here](https://github.com/keepsimple1/mdns-sd/blob/7293e7c5d1bbb7937adabb341a0a388c397b2779/src/service_daemon.rs#L1999) we send out a query to both ipv6 and ipv4 if the ```MyIntf``` instance still exists

keepsimple1 (commented):
Looks good. Only minor comments inline. Thanks for the update!

keepsimple1:
>  ok if we leave this as a follow-up?

Yes, sounds good.

keepsimple1:
One more minor comment: could you mention the issue number that triggered this PR in the Description section?  such that GitHub will automatically link this PR to that issue.

tests/mdns_test.rs
```diff
@@ -689,6 +689,67 @@ fn service_with_ipv4_only() {
     d.shutdown().unwrap();
 }
 
+#[test]
+fn service_with_ipv6_intf() {
+    // Create a daemon
+    let d = ServiceDaemon::new().expect("Failed to create daemon");
+
+    // Register a service with a name len > 15.
```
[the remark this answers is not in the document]
twizansk: yes sorry.  same for the similar comment below.  fixed

tests/mdns_test.rs
```diff
@@ -689,6 +689,67 @@ fn service_with_ipv4_only() {
...
+        None,
+    )
+    .expect("invalid service info")
+    .enable_addr_auto();
+    my_service.set_interfaces(vec![IfKind::IPv6]);
+    my_service.set_link_local_only(true);
+    let result = d.register(my_service);
+    assert!(result.is_ok());
+
+    // Browse for a service and verify all addresses are IPv4.
```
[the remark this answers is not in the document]
twizansk: fixed

twizansk:
> One more minor comment: could you mention the issue number that triggered this PR in the Description section? such that GitHub will automatically link this PR to that issue.

done

keepsimple1 (approved):
LGTM, thanks!

PR #401, opened 2025-10-30
Add `with_timeout` API

### LINKED ISSUE #320: Asynchronous `recv_timeout` and `recv_deadline`
I do not know if it is feasible, but would it be possible to implement a `recv_timeout` API which is `async` and it does not necessarily block the main thread? The same goes for the `recv_deadline` API. All other APIs are `asynchronous` except for those two. Is there a main reason?

Luni-4:
First, sorry for the long delay, but my life has been quite busy lately. 

This PR aims to fix #320, but it also serves as a good foundation for further discussion. Based on the considerations outlined [here](https://github.com/keepsimple1/mdns-sd/issues/320#issuecomment-2691965658), no wrapper was initially considered. However, I’d like to propose my solution regardless. If this PR is not deemed acceptable, feel free to close it at any time. 

I will now walk through the implementation details in a Q&A format.

**Why not futures-utils?**

`futures-utils` seems unmaintained. The most recent release was over a year ago. Additionally, there’s no asynchronous timer in `futures-utils`, meaning an external dependency would still be required, resulting in two dependencies instead of one.

**Why tokio?**

`tokio` is updated daily, includes all the necessary tools, and requires only two features to compile. Therefore, the build process is pretty fast.

**Could there be conflicts if we use a different version of `flume` than the one we re-export?**

Yes, because `with_timeout` uses the same version of `flume` that we re-export, ensuring it remains in sync with future updates.

**Why not `with_deadline` API?**

The `recv_deadline` function relies on `std::time::Instant`, a monotonic clock that behaves differently across operating systems. A monotonic clock measures elapsed time, but does not correspond to a specific point in time, so unusable for an `async` context.  [Here](https://doc.rust-lang.org/std/time/struct.Instant.html) a more detailed description. That is why I chose not to implement the API. Although, we can also use something like `instant.checked_duration_since(Instant::now())` to retrieve a `Duration`, I believe this approach reduces the implementation freedom of a developer.

Additionally, the API is behind the `async` feature, including the tests. Let me know what you think about my idea @keepsimple1, thank you!

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,6 +28,7 @@ tokio = { version = "1.48.0", defalut-features = false, features = ["macros", "t
 
 [dev-dependencies]
 env_logger = { version = "= 0.10.2", default-features = false, features= ["humantime"] }
+tokio = { version = "1.48.0", defalut-features = false, features = ["macros", "rt"]}
 fastrand = "2.3"
 humantime = "2.1"
 test-log = "= 0.2.14"
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -204,3 +204,77 @@ pub async fn with_timeout<T>(
         }
     }
 }
+
+#[cfg(feature = "async")]
+#[cfg(test)]
+// All tokio tests are executed concurrently
+mod tests {
+
+    use core::time::Duration;
+
+    use crate::service_daemon::my_ip_interfaces;
+    use crate::{ServiceDaemon, ServiceEvent, ServiceInfo};
+
+    use super::with_timeout;
+
+    #[tokio::test]
+    async fn test_with_timeout() {
+        let d = ServiceDaemon::new().expect("Failed to create daemon");
+
+        let service = "_test_inval_ptr._udp.local.";
+        let host_name = "my_host_tmp_invalidated_ptr.local.";
+        let intfs: Vec<_> = my_ip_interfaces(false);
+        let intf_ips: Vec<_> = intfs.iter().map(|intf| intf.ip()).collect();
+        let port = 5201;
+        let my_service =
+            ServiceInfo::new(service, "my_instance", host_name, &intf_ips[..], port, None)
+                .expect("invalid service info")
+                .enable_addr_auto();
+        let result = d.register(my_service.clone());
+        assert!(result.is_ok());
+
+        let browse_chan = d.browse(service).unwrap();
+        let timeout = Duration::from_secs(2);
+
+        let mut resolved = false;
+        while let Ok(event) = with_timeout(&browse_chan, timeout).await {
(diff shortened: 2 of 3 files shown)


keepsimple1:
Welcome back @Luni-4 ! 

I appreciated your PR, but unfortunately one of the very reasons I created `mdns-sd` is _not_ to depend on any async runtime, `tokio` included.

Another reason I don't think I can merge this PR is:  I don't think it's in this lib's scope to support async `recv_timeout`. IMO, it should be an enhancement to `flume` crate if one wants to add this API.   (Once `flume` has it, all its users can benefit from it, not just `mdns-sd`.)  And I did a quick vibe coding with Claude Code, it came up with an implementation in `flume` using the similar pattern used by the current `recv_async`, so it seems possible.

If `flume` never would accept such a new API, and you / users of this lib really wants such API,  we could look into defining a `future` that support a timer / deadline on top of it. But I don't know what it would look like, at this moment. Again, I'll still recommend the user to consider use `recv_async` with a timer in their code as the 1st option.

Luni-4:
Thank you, @keepsimple1!

And thank you for your clear response. I was already aware of your position, but since I’ve created something similar for my project—essentially a wrapper over `mdns-sd` using `tokio` as the async runtime—I thought it might still be worth attempting to propose this PR.

If the `flume` developers haven’t added something similar, I believe it’s because they want to remain independent of any specific async runtime, too. That’s just my personal conjecture, based on their [dependencies](https://github.com/zesterer/flume/blob/master/Cargo.toml#L30-L34). Furthermore, `flume` has been in [casual maintenance mode](https://github.com/zesterer/flume?tab=readme-ov-file#status) for the past year.

I have another suggestion. Instead of discarding this PR, how about we add `test_with_timeout` to the `example` directory to demonstrate how to use `mdns-sd` with timeouts in an async context? I believe that would be a valuable use-case. Personally, I don’t think adding `test_with_timeout` as a `doc test` to `lib.rs` would be particularly useful. Let me know what you think of this idea.

keepsimple1:
> I have another suggestion. Instead of discarding this PR, how about we add `test_with_timeout` to the `example` directory to demonstrate how to use `mdns-sd` with timeouts in an async context? I believe that would be a valuable use-case. Personally, I don’t think adding `test_with_timeout` as a `doc test` to `lib.rs` would be particularly useful. Let me know what you think of this idea.

Sorry I don't feel the need for adding such an example. I wanted to keep the examples focused on the core values added by this lib itself.

keepsimple1:
And let me know if you have other questions for this PR. If not, I'm ready (or you can) close it.

Luni-4:
No more questions. Closing this PR.

Issue #413, opened 2026-01-12
Optional reuse port for old devices


Hanssen0:
Hi!

I'm trying to add mDNS service for an old device (specifically, Kindle PW2 from 2013). It runs a Linux 3.0.35 kernel, which doesn't support `SO_REUSEPORT`.

I added a feature flag on my fork and tested that it works, so I plan to open a PR here.

https://github.com/Hanssen0/mdns-sd/commit/3b995b80af5d35f774b547c922d5f02d4abab0f3

Before that, my question is, do you prefer to control this feature manually by devs or silently ignore the error and then continue if it fails?

keepsimple1:
Thanks for opening this!

> do you prefer to control this feature manually by devs or silently ignore the error and then continue if it fails?

I would prefer the devs control this feature manually.

keepsimple1:
PR merged, thanks!

Hanssen0:
Thank you for the great work!

keepsimple1:
FYI, this is released in v0.17.2 🚀

PR #414, opened 2026-01-13
feat: `reuseport` feature

Hanssen0:
Add a default feature `reuseport` to control if `SO_REUSEPORT` should be used. Useful for old Linux kernels (before 3.9).

For more context, see #413


keepsimple1 (approved):
LGTM! Thanks!

Issue #418, opened 2026-01-19
RFC 6763 Non-Compliance: Instance Name Not Escaped in ServiceInfo::new()


thibaut-pascal:
# mDNS Instance Name Escaping Issue

The `mdns-sd` library does not escape special characters (dots `.` and backslashes `\`) in the Instance Name parameter when constructing Service Instance Names, violating RFC 6763 Section 4.3 requirements.

## RFC 6763 Requirements

According to [RFC 6763 Section 4.1.1](https://datatracker.ietf.org/doc/html/rfc6763#section-4.1.1), the DNS Instance name can contain almost any characters, including dots.
```
4.1.1.  Instance Names

   The <Instance> portion of the Service Instance Name is a user-
   friendly name consisting of arbitrary Net-Unicode text [RFC5198].  It
   MUST NOT contain ASCII control characters (byte values 0x00-0x1F and
   0x7F) [RFC20] but otherwise is allowed to contain any characters,
   without restriction, including spaces, uppercase, lowercase,
   punctuation -- including dots -- accented characters, non-Roman text,
   and anything else that may be represented using Net-Unicode.
```

According to [RFC 6763 Section 4.3](https://datatracker.ietf.org/doc/html/rfc6763#section-4.3):

```
4.3.  Internal Handling of Names

   If client software takes the <Instance>, <Service>, and <Domain>
   portions of a Service Instance Name and internally concatenates them
   together into a single string, then because the <Instance> portion is
   allowed to contain any characters, including dots, appropriate
   precautions MUST be taken to ensure that DNS label boundaries are
   properly preserved.  Client software can do this in a variety of
   ways, such as character escaping.

   This document RECOMMENDS that if concatenating the three portions of
   a Service Instance Name, any dots in the <Instance> portion be
   escaped following the customary DNS convention for text files: by
   preceding literal dots with a backslash (so "." becomes "\.").
   Likewise, any backslashes in the <Instance> portion should also be
   escaped by preceding them with a backslash (so "\" becomes "\\").
   Having done this, the three components of the name may be safely
   concatenated.  The backslash-escaping allows literal dots in the name
   (escaped) to be distinguished from label-separator dots (not
   escaped), and the resulting concatenated string may be safely passed
   to standard DNS APIs like res_query(), which will interpret the
   backslash-escaped string as intended.
```

[...]

PR #420, opened 2026-01-20
feat: implement RFC 6763 Section 4.3 escaping for instance names

thibaut-pascal:
Add proper escaping of dots and backslashes in DNS-SD instance names according to RFC 6763 Section 4.3:
- '.' becomes '\.'
- '\' becomes '\\'

This ensures that literal dots in instance names (e.g., "My.Service") are not interpreted as DNS label separators when constructing the Service Instance Name.

Fixes improper handling of instance names containing dots or backslashes.

### DIFF
--- a/src/dns_parser.rs
+++ b/src/dns_parser.rs
@@ -1459,6 +1459,59 @@ impl DnsOutPacket {
         self.size += 2;
     }
 
+    /// Parses a DNS name that may contain escaped characters according to RFC 6763 Section 4.3.
+    /// Returns a vector of labels where each label is the unescaped content.
+    ///
+    /// Escape sequences:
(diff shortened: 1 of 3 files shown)


keepsimple1:
The description section seems to be missing backslash in these: 

- `.` becomes `\.`
- `\` becomes `\\`

I think it's because of Markdown syntax, and should use ``, not '' for them.

keepsimple1:
Another small thing: could you please explicitly refer to the issue number in this PR description?

keepsimple1 (approved):
Overall LGTM, so I will merge this before it's getting too old ;-)  

For reference, this is to fix issue #418 .

PR #423, opened 2026-01-20
Return errors from send_dns_outgoing

thibaut-pascal:
fix of the PR https://github.com/keepsimple1/mdns-sd/pull/419 as I cannot contribute to it, I create a new one


thibaut-pascal:
`test_name_tiebreaking` fails but I'm wondering if this test is really deterministic, there may be a timing issue.

thibaut-pascal:
Discussion continues here https://github.com/keepsimple1/mdns-sd/pull/419

PR #424, opened 2026-01-21
fix: only add scope to unicast link local v6 addresses when printing

hrzlgnm:
The scope id only makes sense in case the address is a v6 uni-cast link local address.

Before the change the output of the query example, facing multiple ipv6 addresses with different scopes would be:

```
At 7.939648ms: Resolved a new service: FritzBoxRepeater._fbox._tcp.local.
 host: FritzBoxRepeater.fritz.box.
 port: 49000
 Address: 192.168.178.89
 Address: fd00::ca0e:14ff:feff:416%enp12s0 <- is an ULA (Unique Local Address, fc00::/7)
 Address: 2003:e8:bf1e:9b00:ca0e:14ff:feff:416%enp12s0 <- is globally scoped
 Address: fe80::ca0e:14ff:feff:416%enp12s0 <- only here it makes sense
 ```
 
 After:
 ```
 At 6.967689ms: Resolved a new service: FritzBoxRepeater._fbox._tcp.local.
 host: FritzBoxRepeater.fritz.box.
 port: 49000
 Address: 2003:e8:bf1e:9b00:ca0e:14ff:feff:416
 Address: fd00::ca0e:14ff:feff:416
 Address: 192.168.178.89
 Address: fe80::ca0e:14ff:feff:416%enp12s0
 ```

### DIFF
--- a/src/dns_parser.rs
+++ b/src/dns_parser.rs
@@ -8,6 +8,7 @@
 use crate::log::trace;
 
 use crate::error::{e_fmt, Error, Result};
+use crate::service_info::is_unicast_link_local;
 
 use if_addrs::Interface;
 
@@ -143,7 +144,7 @@ impl fmt::Display for ScopedIp {
         match self {
             ScopedIp::V4(v4) => write!(f, "{}", v4.addr),
             ScopedIp::V6(v6) => {
-                if v6.scope_id.index != 0 {
+                if v6.scope_id.index != 0 && is_unicast_link_local(&v6.addr) {
                     #[cfg(windows)]
                     {
                         write!(f, "{}%{}", v6.addr, v6.scope_id.index)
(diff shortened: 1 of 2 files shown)


hrzlgnm:
Ah this won't work without updating rust to 1.84.0, perhaps something to be done when bumping the MSRV to 1.84

keepsimple1:
> Ah this won't work without updating rust to 1.84.0, perhaps something to be done when bumping the MSRV to 1.84

Can you use this helper we already have:
https://github.com/keepsimple1/mdns-sd/blob/3f34136122b906c4224d0e2a44b6cc5876740946/src/service_info.rs#L1232-L1237

hrzlgnm:
> > Ah this won't work without updating rust to 1.84.0, perhaps something to be done when bumping the MSRV to 1.84
> 
> Can you use this helper we already have:
> 
> https://github.com/keepsimple1/mdns-sd/blob/3f34136122b906c4224d0e2a44b6cc5876740946/src/service_info.rs#L1232-L1237

Thanks for pointing it out, I totally forgot about that helper.

PR #426, opened 2026-01-27
Try out lower mio version

keepsimple1:
In some environments, `mio` v1.1 conflicts with the user's runtime. Trying out a lower version of `mio`.


keepsimple1:
This was an experiment, no need to merge.

Issue #437, opened 2026-03-06
Getting the local address for a discovered service?


hgomersall:
I'm trying to use `mdns-sd` on a system that has multiple interfaces. Each interface has a link-local IP address. On the networks those interfaces are connected to are various servers all of which have link-local IP addresses, and all of which advertise themselves with mdns-sd.

My understanding is that this is an acceptable design, and that I can specify a connection by binding a socket to the relevant IP address (which defines the interface). The problem I have is that I can't determine the IP address of the interface that a service was discovered on. I guess I could also know what interface, but it's not clear how I can do that either.

Am I missing something here? It seems like a use-case that _should_ be supported. I'm happy to do the leg-work to make this happen, but I need a bit of a steer on whether the info is available already, and failing that, how I should best implement the changes.

For context, I'm seeing an issue on (at least) Windows where connection to a server takes a long time after a quick mdns-sd discovery. Presumably because the machine does not know which interface to use for the connection so has to try the various interfaces until it finds one that works. I suspect it will break horribly if two machines on different physical networks have the same IP address.

keepsimple1:
If I understand it, the issue is that when a ServiceResolved event happens, it contains ResolvedService that does not directly say on which interface the event was triggered.

At the high level, this is due to an early decision I made (without realizing the subtle implications): combine address info (resolved) for the same service instance into one resolved struct. In other words, the resolved service is not per interface, i.e. not interface-scoped. In other words, this lib automatically deduplicates the same service found on multiple interfaces.

Now I found that Apple's Bonjour and Avahi on Linux are not doing that, and are reporting services per interface. I think we could change this lib to match that, but it's probably a quite big change that needs some careful design.

I guess I could also know what interface, but it's not clear how I can do that either.

Today you could find out what interface is if it's IPv6: ScopedIpV6 tells which interface reported the address:

https://github.com/keepsimple1/mdns-sd/blob/e3f4e89ceb3c3f27b6dd45b5b12d17299980abc2/src/dns_parser.rs#L69-L72

But you would have to find out the IP address of this interface via other methods. My search shows that it's possible to bind a socket to interface by using interface index, but I haven't done that before.

For IPv4, currently we don't report the interface index that resolved the address. We can add it for ScopedIpV4 so it's at least available for use. Btw, for IPv4, you can generally bind the socket to 0.0.0.0 and let the OS to select the correct interface, because IPv4 link-local address should be unique between interfaces. (That's not true for IPv6).

[...]

PR #439, opened 2026-03-09
add interface_id in ScopedIpV4

keepsimple1:
Help fixing issue #437 

## Problem

Currently for IPv4 there is no way to find out on which interface the service address(es) are found, while IPv6 doesn't have this problem. 

## Solution
Added `interface_id()` to `ScopedIpV4` to match `ScopedIpV6` that returns the interface on which a service address is found.

### DIFF
--- a/src/dns_parser.rs
+++ b/src/dns_parser.rs
@@ -33,6 +33,18 @@ pub struct InterfaceId {
     pub index: u32,
 }
 
+impl InterfaceId {
+    /// Returns all IP addresses associated with this interface by querying the OS.
+    pub fn get_addrs(&self) -> Vec<IpAddr> {
+        if_addrs::get_if_addrs()
+            .unwrap_or_default()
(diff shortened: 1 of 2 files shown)


hgomersall:
It looks like this would certainly allow a workaround. I think probably by iterating over the interfaces from `if_addrs` and comparing each with the returned `InterfaceId`. That said, is there a reason the address can't be added to the interface ID directly from `if_addrs` to avoid the extra layer?

keepsimple1:
> That said, is there a reason the address can't be added to the interface ID directly from if_addrs to avoid the extra layer?

I didn't want to do that as it feels like against the design purpose of `struct InterfaceId`, but I think we could add a convenient method to do the legwork for the users.  Please see the updated diff for `get_addrs()` method of `InterfaceId`.  Would that work for you?

hgomersall:
Well done, thanks for this!

Issue #438, opened 2026-03-08
Implement Deserialize/Serialize for mdns-sd types


rabbit-time:
Hello. I have found the need to send the data represented by the `ResolvedService` struct (in my case, over a Unix socket) and thought it would be nice if mdns-sd implemented this.

I know dependencies are a concern, but perhaps a 'serde' feature flag would be in order? Seems like to me it wouldn't be too hard to implement and can mostly be done through derive.

FelixSelter:
I could do this if desired by the owner

keepsimple1:
> I know dependencies are a concern, but perhaps a 'serde' feature flag would be in order? Seems like to me it wouldn't be too hard to implement and can mostly be done through derive.

Sure, sounds like a good idea to add this as an optional feature with a default off. 

> I could do this if desired by the owner

That'd be great! Please go ahead. If any questions, please ask in this issue or in your PR.  Thanks for helping!

PR #440, opened 2026-03-09
Serde Deserialize+Serialize implementation

rabbit-time:
This PR introduces a new feature flag: `serde`

I've implemented the `Serialize` and `Deserialize` traits using `derive` for `ResolvedService` and for all the local types which it contains.

Additionally, I've also written a basic test to accompany the changes and added `serde_json` as a development dependency.

If there is a benefit to providing `serde` support for types other than `ResolvedService`, please comment that.

Closes: #438

### DIFF
--- a/src/service_info.rs
+++ b/src/service_info.rs
@@ -1712,8 +1712,8 @@ mod tests {
         use super::{Ipv4Addr, Ipv6Addr};
         use crate::{ResolvedService, ScopedIp, TxtProperties};
 
-        use std::net::IpAddr;
         use std::collections::HashSet;
+        use std::net::IpAddr;
 
         #[test]
         fn test_deserialize_serialize() -> Result<(), Box<dyn std::error::Error>> {


rootrascal:
CI jobs are currently failing since clippy isn't being run with features enabled.

Line 1286 in `./src/service_info.rs`:
`#[cfg_attr(feature = "serde", derive(Deserialize, Serialize))]`

src/service_info.rs
```diff
@@ -1697,4 +1708,53 @@ mod tests {
         assert!(!service_info.is_address_supported(&intf_loopback_v4));
         assert!(service_info.is_address_supported(&intf_loopback_v6));
     }
+
+    #[test]
```
keepsimple1: You can add `#[cfg(feature = "serde")]` here, which would also fix the CI error, I think.

keepsimple1 (commented):
Looks good! One inline comment related to the CI error.  And optional, if you have bandwidth, you can also add a new step in CI to run `cargo clippy --tests --all-features -- -D warnings`  to cover the features (I think).

rootrascal:
Also, one small thing to note. Currently, when there is no `scope_id` on an IPv6 address, it just puts in blank values. 

@keepsimple1 My question to you is, should this be considered intended behavior as it reflects the internal representation with how the struct is implemented in Rust (uses default empty values as opposed to `Option<T>`) or should it be optimized for size and omit these fields in serialization?

image

keepsimple1:
> should this be considered intended behavior as it reflects the internal representation with how the struct is implemented in Rust (uses default empty values as opposed to Option<T>) or should it be optimized for size and omit these fields in serialization?

I think we should keep it as is because `scope_id` is not Option in `ScopedIpV6`.  It's empty here is because it's manually constructed in a test. It would not be empty in a real world service.  (Or did I miss something?)

Btw, now `ScopedIpV4` also got `interface_id`, which is the same as `scope_id` for IPv6. (see merged PR #439 ).

rootrascal:
> I think we should keep it as is because `scope_id` is not Option in `ScopedIpV6`. It's empty here is because it's manually constructed in a test. It would not be empty in a real world service. (Or did I miss something?)
> 
> Btw, now `ScopedIpV4` also got `interface_id`, which is the same as `scope_id` for IPv6. (see merged PR #439 ).

To my knowledge, an IPv6 address is not always accompanied by a scope ID. I may be wrong.

---

I will work on finalizing this PR today.

rootrascal:
Alright everything should be ready to go!

If you'd like, take one last look over everything. I moved the serde test into its own submodule so clippy would stop complaining about unused imports.

src/service_info.rs
```diff
@@ -1697,4 +1705,63 @@ mod tests {
...
+                  "addr": "fe80:2001:db8:85a3:0:8a2e:370:7334",
+                  "scope_id": {
+                    "name": "",
+                    "index": 0
+                  }
+                },
+                {
+                  "addr": "127.0.0.1"
+                }
+            ],
```
keepsimple1: `addresses` is HashSet, and its iteration order that is non-deterministic. It's possible this check against of serialized JSON text pass on some machines but not on others.

Can we deserialize the json back and compare structurally with the original?  As a bonus it  covers the deserialization path. Or maybe compare parsed `serde_json::value` objects instead of text?

PR #444, opened 2026-03-21
refactoring: ScopedIpV4 to use multiple InterfaceIds

keepsimple1:
### Problem

`ScopedIpV4` previously had a single `interface_id` (added in #439). This meant the same IPv4 address discovered via different interfaces would appear as two distinct entries in ResolvedService.addresses. This can cause confusion about how many IPv4 addresses actually discovered.

Unlike IPv6 link-local addresses where the scope ID is part of the address identity, for IPv4 the interface is discovery metadata.

### Changes

- `ScopedIpV4` to hold `Vec<InterfaceId>` instead of a single `interface_id`.
- Implements `Eq/Hash` manually based on `addr` only, so the same IPv4 address deduplicates correctly in HashSet<ScopedIp>.

### DIFF
--- a/src/dns_parser.rs
+++ b/src/dns_parser.rs
@@ -21,6 +21,7 @@ use std::{
     collections::HashMap,
     convert::TryInto,
     fmt,
+    hash::{Hash, Hasher},
     net::{IpAddr, Ipv4Addr, Ipv6Addr},
     str,
     time::SystemTime,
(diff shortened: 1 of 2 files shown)


keepsimple1:
@hgomersall @hrzlgnm May I ask your thoughts about this PR?

hrzlgnm:
> @hgomersall @hrzlgnm May I ask your thoughts about this PR?

Looks reasonable to me, looking forward in using the addition 👍🏻

hgomersall:
The point of the change then is to combine a single service onto multiple interfaces? That is, if a service is discovered on multiple interfaces they are all added to the interfaces scope?

How are services uniquified? What would happen if different services with the same name are advertised on genuinely different networks and just happen to have the same (link-local) IP address?

FWIW, I currently append a UUID of the service to the txt fields and check as I'm not aware of a way this can be handled without some meta-checking.

hgomersall:
If my understanding is correct, I even wonder whether there should be a closure API somewhere that handles the uniquification of a service.

keepsimple1:
> The point of the change then is to combine a single service onto multiple interfaces?

This change is not to combine service(s).

The change is to combine multiple entries in `ResolvedService.addresses` that have the same IP but different `InterfaceId` into a single entry. 

> How are services uniquified? What would happen if different services with the same name are advertised on genuinely different networks and just happen to have the same (link-local) IP address?

AFAIK, In IPv4, on one host, the link-local IP address cannot be the same on different networks via different interfaces. This is different from IPv6.  (If I was wrong, would you share a use case of such?  I am genuinely interested).  

That said, this PR is a refactoring. It's not trying to solve the longer term question of reporting services per interface. (I have some idea recently, will sort it out in a different ticket).

hgomersall:
> AFAIK, In IPv4, on one host, the link-local IP address cannot be the same on different networks via different interfaces. This is different from IPv6. (If I was wrong, would you share a use case of such? I am genuinely interested).

It's more that the service itself could have the same IP as another service on a different interface. It those services are on different networks they can have the same IP address and potentially the same mdns name.

keepsimple1:
> It's more that the service itself could have the same IP as another service on a different interface. It those services are on different networks they can have the same IP address and potentially the same mdns name.

I think I will be addressed in a separate PR.  Merging the current one for now.

hgomersall:
FWIW, the previous variant of this I have tested and it seems to work well for my use case. I'll move on to this variant shortly.

PR #454, opened 2026-04-05
chore(deps): update

CosminPerRam:
Also removes `test-log-macros` as we are on `1.71` [now](https://github.com/keepsimple1/mdns-sd/pull/324#issuecomment-2708655869).
Not sure why we had `fastrand` defined again in dev-deps (as its already in deps), #455 removes its usage in tests.

### DIFF
--- a/tests/mdns_test.rs
+++ b/tests/mdns_test.rs
@@ -1515,6 +1515,7 @@ fn test_cache_flush_record() {
     .unwrap();
     let result = server.register(my_service);
     assert!(result.is_ok());
+    assert!(false);
 
     timed_println(format!(
         "Re-registered with updated IPv4 addr: {}",
@@ -2575,7 +2576,6 @@ fn test_interface_id_get_addrs() {
 
 /// A helper function to include a timestamp for println.
 fn timed_println(msg: String) {
-    let now = SystemTime::now();
-    let formatted_time = humantime::format_rfc3339(now);
+    let formatted_time = jiff::Timestamp::now().to_string();
     println!("[{}] {}", formatted_time, msg);
 }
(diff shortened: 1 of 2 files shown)


tests/mdns_test.rs
```diff
@@ -2575,7 +2575,6 @@ fn test_interface_id_get_addrs() {
 
 /// A helper function to include a timestamp for println.
 fn timed_println(msg: String) {
-    let now = SystemTime::now();
-    let formatted_time = humantime::format_rfc3339(now);
+    let formatted_time = jiff::Timestamp::now().to_string();
```
CosminPerRam, line 2578: This produces
`2026-04-05T19:18:51.276776100Z`
instead of before
`2026-04-05T19:18:51.276776Z`
... which I don't think its a problem necessarily.

keepsimple1: Like you mentioned, it's not a problem. I.e. I don't think we need to introduce `jiff`, which brings in more dependencies. `humantime` has no dependency.

CosminPerRam: I swapped `humantime` with `jiff` because its being used in the latest release of [env_logger](https://github.com/rust-cli/env_logger/blob/92f8d8d08343c30e60b5f54455d7e16c810fcf11/Cargo.toml#L130), so its actually one less dependency.

keepsimple1: Got it. Since `env_logger` is for testing only, I think it's OK for it staying with the current version.  Can you please remove the update for `env_logger`, and hence no need to add `jiff` ?  

The rest LGTM. Thanks!

CosminPerRam: The replacement of `humantime` has been done in [0.11.7](https://github.com/rust-cli/env_logger/releases/tag/v0.11.7), so I've pinned the version to `0.11.6`, please mention if I should roll it back completely anyways.

keepsimple1 (approved):
LGTM, thanks!

PR #455, opened 2026-04-05
tests: remove explicit random port assignment

CosminPerRam:
Explicitly assigning 'random' ports could fail, instead, let the host decide on it.


keepsimple1 (approved):
LGTM

Issue #459, opened 2026-04-16
Multiple interface address case


anti-social:
I have a service with 2 addresses on an interface (`192.168.1.148/24` and `10.238.0.51/24`) and a discoverer with only one address (`192.168.1.2/24`). With this configuration the discoverer does not always discover the service. In `wireshark` I can see that sometimes the service chooses `10.238.0.51` ip address to communicate (as addresses are stored in a `HashSet` it depends on its seed). Thus the discoverer does not see these packets as it does not have such an address.

I think it would be worth sending dedicated announcements for every address on an interface and responses only for matched addresses with a source one. I have a quick patch which implements this behavior although it is a little hacky.

keepsimple1:
That's indeed a problem.  I've created a PR #460 to fix that, could you please give a try?  Please feel free to comment on the PR.

anti-social:
What do you think about an option to filter out resolved addresses to return only ones that are matched to the interface networks? Something like:

```rust
service_daemon.same_subnet_only(true);
```

Why it would be nice to have this option. To do the same thing in the application we will have to add additional dependencies to scan network interfaces to be able to filter out unsolicited addresses.

Either we could add some flag to the `ScopedIpV4` and `ScopedIpV6` structs.

keepsimple1:
Instead of a daemon-wide flag like `service_daemon.same_subnet_only`, maybe we can support an option to filter per-browse? Something like:

```rust
pub fn browse_with(&self, service_type: &str, options: BrowseOptions)
    -> Result<Receiver<ServiceEvent>>;

pub struct BrowseOptions {
    pub same_subnet_only: bool,
    // pub cache_only: bool,  (then we can deprecate browse_cache)
    // room to grow
}
```
The rationale is that different queriers might want different policies for whatever reason (say, a particular service type).

It would need some plumbing to filter out unwanted addresses when sending back `ServiceResolved` event, but I think it's doable.  And if addresses were all filtered out, the event will not be sent. 

And I think such API could be further extended to support other features in the future.  What do you think?

anti-social:
Yep, I like it. Then I'll try to make a PR

PR #460, opened 2026-04-17
fix: use a source IP matching the querier's subnet when responding

keepsimple1:
This is to fix issue #459 .

### DIFF
--- a/src/service_daemon.rs
+++ b/src/service_daemon.rs
@@ -4660,7 +4660,12 @@ mod tests {
         },
         service_daemon::{add_answer_of_service, check_hostname},
     };
-    use std::time::{Duration, SystemTime};
+    use if_addrs::{IfAddr, Ifv4Addr};
+    use std::{
+        collections::HashSet,
+        net::{IpAddr, Ipv4Addr},
+        time::{Duration, SystemTime},
+    };
     use test_log::test;
 
     #[test]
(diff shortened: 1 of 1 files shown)


anti-social:
We also should send announcements from every interface address. One more problem that even replying from right address we send all the interface addresses so other side should filter them. I believe the library must do the work either on a service side or a discoverer.

Here is my patch (for me it works but obviously there are some corner cases that should be fixed): https://github.com/anti-social/mdns-sd/commit/e721f757123d82aaa57639e3ec08b03f9e12b713

keepsimple1:
> We also should send announcements from every interface address. 

That's a good point. At the multicast (L2) layer, the listener should receive the announcement packets regardless the source address, but the IP layer might drop them.  It's essentially same problem as the response case of this ticket. 

> One more problem that even replying from right address we send all the interface addresses so other side should filter them. 

I'm a bit worried about this. I cannot find a direct link, but I recall there has been some cases the publisher wants to include some address records that's not in the subnet, sometime even include the loopback address.

> Here is my patch

I'll take a closer look. Please feel free to convert it into a PR. (I can close mine later.) But maybe fix the response (and the announcement) case first, and deal with the addresses records question later in a separate PR?

anti-social:
Possibly it is worth making an option for a discoverer to filter out unmatched IPs. In my case I show a list of found services to a user and it is annoying that there are addresses which are not connectable.

Let's merge this PR and I'll try to create another one for announcements and IP filtering.

keepsimple1:
Sounds good. Merged this one. Please go ahead.

keepsimple1:
> Let's merge this PR and I'll try to create another one for announcements and IP filtering.

just checking if you still plan to work on that, or if need anything from me.

anti-social:
> just checking if you still plan to work on that, or if need anything from me.

It turned out a little trickier than I expected. Here it is: #462 . Tomorrow I'll try to glimpse at it one more time.

PR #462, opened 2026-04-23
Send outgoing packets for every address

### LINKED ISSUE #459: Multiple interface address case

anti-social:
Partially fixes #459

### DIFF
--- a/src/dns_parser.rs
+++ b/src/dns_parser.rs
@@ -1416,13 +1416,14 @@ impl DnsRecordExt for DnsNSec {
     }
 }
 
-#[derive(PartialEq)]
+#[derive(Debug, PartialEq)]
 enum PacketState {
     Init = 0,
     Finished = 1,
 }
 
 /// A single packet for outgoing DNS message.
+#[derive(Debug)]
 pub struct DnsOutPacket {
     /// All bytes in `data` is the actual packet on the wire.
     data: Vec<u8>,
@@ -1916,7 +1917,7 @@ impl DnsOutgoing {
     }
 
     /// Returns a list of actual DNS packet data to be sent on the wire.
-    pub fn to_data_on_wire(&self) -> Vec<Vec<u8>> {
+    pub(crate) fn _to_data_on_wire(&self) -> Vec<Vec<u8>> {
         let packet_list = self.to_packets();
         packet_list.into_iter().map(|p| p.data).collect()
     }
--- a/src/dns_parser.rs
+++ b/src/dns_parser.rs
@@ -1916,12 +1916,6 @@ impl DnsOutgoing {
         self.questions.push(q);
     }
 
-    /// Returns a list of actual DNS packet data to be sent on the wire.
-    pub(crate) fn _to_data_on_wire(&self) -> Vec<Vec<u8>> {
-        let packet_list = self.to_packets();
-        packet_list.into_iter().map(|p| p.data).collect()
-    }
-
     /// Encode self into one or more packets.
     pub fn to_packets(&self) -> Vec<DnsOutPacket> {
         let mut packet_list = Vec::new();
--- a/src/service_daemon.rs
+++ b/src/service_daemon.rs
@@ -34,8 +34,8 @@ use crate::{
     current_time_millis,
     dns_cache::{DnsCache, IpType},
     dns_parser::{
-        ip_address_rr_type, DnsAddress, DnsEntryExt, DnsIncoming, DnsOutgoing, DnsPointer,
-        DnsRecordBox, DnsRecordExt, DnsSrv, DnsTxt, InterfaceId, RRType, ScopedIp,
+        ip_address_rr_type, DnsAddress, DnsEntryExt, DnsIncoming, DnsOutPacket, DnsOutgoing,
+        DnsPointer, DnsRecordBox, DnsRecordExt, DnsSrv, DnsTxt, InterfaceId, RRType, ScopedIp,
         CLASS_CACHE_FLUSH, CLASS_IN, FLAGS_AA, FLAGS_QR_QUERY, FLAGS_QR_RESPONSE, MAX_MSG_ABSOLUTE,
     },
     error::{e_fmt, Error, Result},
@@ -798,7 +798,7 @@ fn _new_socket_bind(intf: &Interface, should_loop: bool) -> Result<MyUdpSocket> 
 
             // Test if we can send packets successfully.
             let multicast_addr = SocketAddrV4::new(GROUP_ADDR_V4, MDNS_PORT).into();
-            let test_packets = DnsOutgoing::new(0).to_data_on_wire();
+            let test_packets = DnsOutgoing::new(0)._to_data_on_wire();
             for packet in test_packets {
                 sock.send_to(&packet, &multicast_addr)
                     .map_err(|e| e_fmt!("send multicast packet on addr {}: {}", ip, e))?;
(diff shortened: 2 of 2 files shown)


keepsimple1:
Thanks for submitting the PR! I understand it's still in Draft, but I have a general comment:  can we split this PR into smaller ones that are relatively independent (or built on-top) of each other?  That will make review and merge much easier. 

For example, those cleanup changes can be in one PR:

- `println!` → `debug!` in `DnsHostInfo::write`.
- u32-overflow fix in `get_expiration_time`.
- add Debug derives on `DnsOutPacket`, etc.

anti-social:
I've fixed u32 overflow in a PR that optimizes dns packet serialization because I caught it in tests that were written for the PR

#465 #466 #467

src/service_daemon.rs
```diff
@@ -3921,8 +3908,8 @@ enum Command {
     /// Announce again a service to local network
     RegisterResend(String, u32), // (fullname)
 
-    /// Resend unregister packet.
-    UnregisterResend(Vec<u8>, u32, bool), // (packet content, if_index, is_ipv4)
+    /// Resend unregister packets.
+    UnregisterResend(Vec<DnsOutPacket>, u32, bool), // (packets to resend, if_index, is_ipv4)
```
anti-social, line 4008: We changed this variant thus `DnsOutPacket` should implement `std::fmt::Debug`

src/service_daemon.rs
```diff
@@ -4237,51 +4224,74 @@ fn is_apple_p2p_by_name(name: &str) -> bool {
...
+        my_intf
+            .addrs
+            .iter()
+            .filter(|addr| addr.ip().is_ipv6())
+            .collect()
     };
-
-    send_dns_outgoing_impl(out, if_name, my_intf.index, if_addr, sock, port)
+    let mut invalid_interfaces = HashSet::new();
+    let packets = out.to_packets();
```
keepsimple1, line 4272: It seems it would  return `packets` even when no address was actually sent on.

src/service_daemon.rs
```diff
@@ -2313,16 +2306,9 @@ impl Zeroconf {
...
-            Ok(sent_vec) => sent_vec,
-            Err(InternalError::IntfAddrInvalid(intf_addr)) => {
-                let invalid_intf_addrs = HashSet::from([intf_addr]);
-                let _ = self.send_cmd_to_self(Command::InvalidIntfAddrs(invalid_intf_addrs));
-                vec![]
-            }
-        };
-        sent_vec.into_iter().next().unwrap_or_default()
+        let (sent_packets, invalid_intf_addrs) = send_dns_outgoing(&out, intf, sock, self.port);
+        let _ = self.send_cmd_to_self(Command::InvalidIntfAddrs(invalid_intf_addrs));
```
keepsimple1: This sends the `InvalidIntfAddrs` command even if `invalid_intf_addrs` is empty.

src/service_daemon.rs
```diff
@@ -4553,7 +4563,7 @@ fn announce_service_on_intf(
 ) -> MyResult<bool> {
     let is_ipv4 = sock.domain() == Domain::IPV4;
     if let Some(out) = prepare_announce(info, intf, dns_registry, is_ipv4) {
-        let _ = send_dns_outgoing(&out, intf, sock, port, None)?;
+        let _ = send_dns_outgoing(&out, intf, sock, port);
```
keepsimple1, line 4573: This call used to propagate the error back to the caller. Now it silently ignores the error.

src/dns_parser.rs
```diff
@@ -1916,7 +1917,7 @@ impl DnsOutgoing {
     }
 
     /// Returns a list of actual DNS packet data to be sent on the wire.
-    pub fn to_data_on_wire(&self) -> Vec<Vec<u8>> {
+    pub(crate) fn _to_data_on_wire(&self) -> Vec<Vec<u8>> {
```
keepsimple1: This change seems odd as it's still used.  If you removed it completely and replace the call sites with direct alternative calls, I'm ok with that.

keepsimple1 (commented):
As this is somewhat non-trivial change, could you please add a detailed PR description:  
- What's the problem? (motivation)
- Changes (approach or the solution)

That will also help the reader understand which part of #459 is addressed by this PR.

And could you please also add new test case to cover the multi-address-per-interface case that motivated this change? 

Thanks!

keepsimple1:
@anti-social friendly bump: will you be able to move forward with this PR?

src/service_daemon.rs
anti-social: As I can see we need a little deeper refactoring to track status for pair of an interface and an address.

keepsimple1: Maybe we could change this function's return type to match the new return type of `send_dns_outgoing` so that it can bubble it up? 

Something like this? 
```rust
fn announce_service_on_intf(
    dns_registry: &mut DnsRegistry,
    info: &ServiceInfo,
    intf: &MyIntf,
    sock: &PktInfoUdpSocket,
    port: u16,
) -> (bool, HashSet<Interface>) {
    let is_ipv4 = sock.domain() == Domain::IPV4;
    if let Some(out) = prepare_announce(info, intf, dns_registry, is_ipv4) {
        let (sent_packets, invalid) = send_dns_outgoing(&out, intf, sock, port);
        return (!sent_packets.is_empty(), invalid);
    }
    (false, HashSet::new())
}
```

The call sites need to change to handle a HashSet but I think doable?

src/service_daemon.rs
```diff
@@ -3265,18 +3239,26 @@ impl Zeroconf {
...
-            }
-
-            let if_name = intf.name.clone();
+                let if_name = intf.name.clone();
 
-            self.increase_counter(Counter::Respond, 1);
-            self.notify_monitors(DaemonEvent::Respond(if_name));
+                self.increase_counter(Counter::Respond, 1);
+                self.notify_monitors(DaemonEvent::Respond(if_name));
+            }
```
keepsimple1, line 3261: I think we should have a fallback to fanout when `matched_source` is None (current diff silently drops responses in such case). 

something like
```rust
else {
    // No matched source — fan out across all addresses on the interface so
    // the querier may hear us from one subnet.
    let (_, invalid) = send_dns_outgoing(&out, intf, &sock.pktinfo, self.port);
    if !invalid.is_empty() {
        let _ = self.send_cmd_to_self(Command::InvalidIntfAddrs(invalid));
    }
}
```

Also this branch now needs a rebase (sorry, it's a bit more work) as `unicast_dest` was added through `send_dns_outgoing`, and the fanout logic needs to be gated on `unicast_dest.is_none()`.

and if we could do that, we can move the counter update out after `if..else..` block, i.e. same as before 

```rust
let if_name = intf.name.clone();
self.increase_counter(Counter::Respond, 1);
self.notify_monitors(DaemonEvent::Respond(if_name));
```

PR #464, opened 2026-04-25
docs: add doc comments for error handling on major APIs

### LINKED ISSUE #463: [DOCS] Describe error cases in common functions
Common functions such as `ServiceDaemon::new` (and `new_with_port` by extension) and `ServiceDaemon::browse` all return a result type but the cases in which they can error are not properly documented.

I would like to suggest adding a `# Errors` section to at least the most commonly used functions in this crate just to define when the errors occur, why, and which error variants are returned. Documenting this is important for proper error handling on the user side of this library.

Reading through the code for `ServiceDaemon`'s constructors, it's not exactly clear to me what (if any) are failure points that users of `mdns-sd` like me should realistically worry about.

keepsimple1:
This is to fix #463 

- Added doc comments about errors in main APIs
- Added a new error variant `DaemonShutdown` for a specific error.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -4,8 +4,19 @@ use std::fmt;
 #[derive(Clone, Debug, PartialEq, Eq)]
 #[non_exhaustive]
 pub enum Error {
-    /// Like a classic EAGAIN. The receiver should retry.
+    /// Like a classic `EAGAIN`. Returned by [`ServiceDaemon`](crate::ServiceDaemon)
+    /// methods when the daemon's bounded command queue is temporarily full,
+    /// so the command could not be enqueued. The caller should retry after a
+    /// short delay; this is not an indication that anything is broken.
     Again,
+
+    /// The daemon thread has exited and its command channel is closed, so the
+    /// command could not be delivered. Returned by [`ServiceDaemon`](crate::ServiceDaemon)
+    /// methods after [`shutdown`](crate::ServiceDaemon::shutdown) has been
+    /// called or after the daemon thread has terminated for another reason.
+    /// Retrying will not help; callers should log and move on (or create a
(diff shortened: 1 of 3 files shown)


rootrascal:
Looks great! If you think it'd be helpful, I would also suggest adding an example for `ServiceDaemon::new` that demonstrates a robust use of the API for an application that relies on it.

```rust
use std::thread;

let mdns = ServiceDaemon::new().or_else(|err| {
    match err {
        Error::Again => {
            thread::sleep(Duration::from_secs(3));
            ServiceDaemon::new()
        }
        _ => Err(err),
    }
})?;
```
Maybe something like this? Or perhaps using recursion.

keepsimple1:
Thanks!  I hear you. For `ServiceDaemon::new()` there is no `Error::Again` error though.  Like in the new docs added, it just returns OS level failures. 

As example, currently we have some basic ones in the overall doc [usage](https://docs.rs/mdns-sd/latest/mdns_sd/#usage), and also in [example](https://github.com/keepsimple1/mdns-sd/tree/main/examples) code.  Are they helpful, or are you looking for example more specific for error handling?   

I updated one line in `lib.rs` examples to highlight `monitor()`.

rootrascal:
Ah, I see. I must have confused it with the `browse` method when reading through the diff which it might be useful for.

Other than that I think the PR is good how it is now.

keepsimple1:
Just to clarify, I didn't add example code of handling `Error::Again` in `browse` doc comments as it seems not many users in the real world do retry for `browse` (almost never has the channel full when starting to browse). 

It seems some users handle `Error::Again` around `shutdown()` or `unregister()` which kind of makes sense.

PR #467, opened 2026-04-27
Optimize dns outgoing serialization

anti-social:


src/dns_parser.rs
```diff
@@ -1440,19 +1437,18 @@ pub struct DnsOutPacket {
 impl DnsOutPacket {
     fn new() -> Self {
         Self {
-            data: Vec::new(),
-            size: MSG_HEADER_LEN, // Header is mandatory.
+            data: vec![0; MSG_HEADER_LEN],
```
anti-social, line 1440: Possibly we could create a vector with capacity of 32 bytes to reduce future reallocations.

src/dns_parser.rs
```diff
@@ -1440,19 +1437,18 @@ pub struct DnsOutPacket {
...
     }
 
     pub fn size(&self) -> usize {
-        self.size
+        self.data.len()
     }
 
-    pub fn to_bytes(&self) -> Vec<u8> {
-        self.data.concat()
+    pub fn as_bytes(&self) -> &[u8] {
```
anti-social, line 1450: If we need backward compatibility I will add `to_bytes(&self) -> Vec<u8>` method

src/dns_parser.rs
```diff
@@ -1935,7 +1918,7 @@ impl DnsOutgoing {
     /// Returns a list of actual DNS packet data to be sent on the wire.
     pub fn to_data_on_wire(&self) -> Vec<Vec<u8>> {
         let packet_list = self.to_packets();
-        packet_list.iter().map(|p| p.data.concat()).collect()
+        packet_list.iter().map(|p| p.data.to_vec()).collect()
```
keepsimple1:
```suggestion
        packet_list.into_iter().map(|p| p.data).collect()
```
nitpick: `packet_list` is owned, so no more need to call `to_vec()`

src/dns_parser.rs
keepsimple1: It's internal to the crate, no need to add that.

src/dns_parser.rs
keepsimple1: I think your current change is OK.  Between 12 and 32 bytes, there is little difference as the final packet is probably 200-300 bytes (or more). 

Preallocating 512 might make a difference but I don't see urgent need unless we have measurement or use case to back it up.

src/dns_parser.rs
```diff
@@ -2601,5 +2584,176 @@ const fn u32_from_be_slice(s: &[u8]) -> u32 {
 const fn get_expiration_time(created: u64, ttl: u32, percent: u32) -> u64 {
     // 'created' is in millis, 'ttl' is in seconds, hence:
     // ttl * 1000 * (percent / 100) => ttl * percent * 10
-    created + (ttl * percent * 10) as u64
+    created + (ttl as u64 * percent as u64 * 10)
```
keepsimple1, line 2587: nice catch!

keepsimple1 (approved):
LGTM! Nice work!  only minor comments inline.

PR #469, opened 2026-05-08
Add support for RFC 6762 legacy unicast responses

luqs1:
_Hi, thanks for your package. I just wanted to contribute some code that my clanker wrote up. It helped me get mdns working when hosted on my macbook from my pixel 8. The rest of this PR is AI generated but hopefully still useful to read haha._

Browsers on Android can't resolve `.local` names published by mdns-sd. They resolve names published by Apple's mDNSResponder fine, and an NsdManager-based app on the same phone sees the records mdns-sd is publishing — so the records are reaching the phone, the browser just isn't getting them.

The reason is that Android's `getaddrinfo` (Mainline DnsResolver, since November 2021) issues mDNS queries from an ephemeral source port and waits for a unicast reply on that port. This is the "legacy/one-shot" case described in [RFC 6762 §6.7](https://datatracker.ietf.org/doc/html/rfc6762#section-6.7). mdns-sd today always replies via multicast from port 5353 — so the answer goes to `224.0.0.251:5353`, which the ephemeral `getaddrinfo` socket isn't listening on, and the lookup times out. The author's own comment at `service_daemon.rs:790` flags this:

```rust
// Here we set the TTL to 255 for multicast as we don't support unicast yet.
```

Same shape on the wire. Before this patch, with my Mac publishing `knock-knock.local`:

```
phone fe80::bc35:….50742 > ff02::fb.5353: A (QM)? knock-knock.local.
mac   fe80::186a:….5353  > ff02::fb.5353: 4/0/0 knock-knock.local. AAAA …  ← phone never sees this
```

After:

```
mac   fe80::186a:….5353  > fe80::bc35:….50742: 1/0/0 A 192.168.1.247  ← unicast back to the ephemeral port
```

Chrome, Firefox, and DuckDuckGo on Android 16 all load `http://knock-knock.local:8000/` immediately after the patch. macOS and iOS were never broken — their resolvers go through Bonjour, not the legacy unicast path.

### What's in the diff

Three logically-distinct changes, all in `service_daemon.rs` plus a small helper in `dns_parser.rs`. Happy to split into separate PRs if you'd rather land them independently — they're somewhat orthogonal but all needed for the Android case.

**Routing the response unicast for legacy queriers.** `pktinfo.addr_src` already had the full `SocketAddr` of the querier; we were calling `.ip()` on it and dropping the port. Plumb the full address through `handle_query`, and when the source port isn't 5353, send the response unicast to it instead of multicasting. New `unicast_on_intf` helper mirrors `multicast_on_intf` for the unicast path. All other callers of `send_dns_outgoing*` pass `None` for the new parameter and behave exactly as before.

**Echoing the question and clearing the cache-flush bit on legacy responses.** §6.7 plus §10.2 — strict legacy resolvers can drop responses that omit the original question section or set the unique bit. Added `DnsOutgoing::clear_cache_flush_bits()` to walk answers, additionals, and authorities and reset.

**Picking the address set by question type, not socket family.** Independent bug found while debugging the above. Today the address set is selected from `is_ipv4 = querier_ip.is_ipv4()` — i.e. based on which socket received the question. So an `A` question received over IPv6 transport gets answered with `AAAA` records (correctly tagged by `ip_address_rr_type`, but the wrong record type for the question that was asked). Android's `getaddrinfo` happily sends both `A` and `AAAA` queries over its preferred IPv6 mDNS socket, so this was the second reason resolution didn't complete after the unicast fix landed. Now `RRType::A` returns IPv4 addresses, `RRType::AAAA` returns IPv6 addresses, `RRType::ANY` returns both, regardless of which socket the question came in on.

While I was in there I also flipped a `return` to `continue` on the empty-addrs branch — the previous code bailed out of the entire `handle_query` if any single A/AAAA question on the interface had no matching addresses, which silently dropped answers for sibling questions in the same query.

### What I didn't change

- The response TTL is unchanged. RFC §6.7 recommends ≤10s for legacy unicast; in practice modern resolvers don't seem to care, and Android accepts the existing default fine. Easy enough to add later behind the same `unicast_dest.is_some()` branch.
- The QU bit on questions is still being masked away in `dns_parser.rs:281-284` as cache-flush. Honoring QU is a stricter form of the same problem and is orthogonal — it only matters for queriers sourcing from port 5353 that explicitly request unicast (see also draft #190, which addressed the QU case but not the legacy port case). Could be a follow-up.
- I have not added unit tests. The behavior is cleanly observable on the wire — open an ephemeral UDP socket, send a question, assert the reply arrives unicast on that socket — happy to write that if you'd like before merge.

### Testing

Reproduced and verified on an Android 16 Pixel 8 against a macOS 15.x host running this branch, with a small axum HTTP server bound to `0.0.0.0:8000` whose mDNS hostname is published as `knock-knock.local.` via `ServiceInfo::new(...).enable_addr_auto()`.

- Before: Chrome / Firefox / DuckDuckGo time out on `http://knock-knock.local:8000/`. `NsdManager` apps see the service. macOS `dns-sd -q knock-knock.local` resolves immediately.
- After: all three browsers load on first attempt. macOS and iOS unaffected. tcpdump confirms responses go unicast to the querier's source port for legacy queriers and stay multicast for proper mDNS clients.

### DIFF
--- a/src/service_daemon.rs
+++ b/src/service_daemon.rs
@@ -2144,14 +2144,14 @@ impl Zeroconf {
                 trace!("sending out probing of questions: {:?}", out.questions());
                 if let Some(sock) = self.ipv4_sock.as_mut() {
                     if let Err(InternalError::IntfAddrInvalid(intf_addr)) =
-                        send_dns_outgoing(&out, intf, &sock.pktinfo, self.port, None)
+                        send_dns_outgoing(&out, intf, &sock.pktinfo, self.port, None, None)
                     {
                         invalid_intf_addrs.insert(intf_addr);
                     }
                 }
                 if let Some(sock) = self.ipv6_sock.as_mut() {
                     if let Err(InternalError::IntfAddrInvalid(intf_addr)) =
-                        send_dns_outgoing(&out, intf, &sock.pktinfo, self.port, None)
+                        send_dns_outgoing(&out, intf, &sock.pktinfo, self.port, None, None)
                     {
                         invalid_intf_addrs.insert(intf_addr);
                     }
@@ -2314,7 +2314,7 @@ impl Zeroconf {
         }
 
         // Only (at most) one packet is expected to be sent out.
-        let sent_vec = match send_dns_outgoing(&out, intf, sock, self.port, None) {
+        let sent_vec = match send_dns_outgoing(&out, intf, sock, self.port, None, None) {
             Ok(sent_vec) => sent_vec,
             Err(InternalError::IntfAddrInvalid(intf_addr)) => {
                 let invalid_intf_addrs = HashSet::from([intf_addr]);
@@ -2358,14 +2358,14 @@ impl Zeroconf {
         let mut invalid_intf_addrs = HashSet::new();
         if let Some(sock) = self.ipv4_sock.as_ref() {
             if let Err(InternalError::IntfAddrInvalid(intf_addr)) =
-                send_dns_outgoing(&out, intf, &sock.pktinfo, self.port, None)
+                send_dns_outgoing(&out, intf, &sock.pktinfo, self.port, None, None)
             {
                 invalid_intf_addrs.insert(intf_addr);
             }
         }
         if let Some(sock) = self.ipv6_sock.as_ref() {
             if let Err(InternalError::IntfAddrInvalid(intf_addr)) =
-                send_dns_outgoing(&out, intf, &sock.pktinfo, self.port, None)
+                send_dns_outgoing(&out, intf, &sock.pktinfo, self.port, None, None)
             {
                 invalid_intf_addrs.insert(intf_addr);
             }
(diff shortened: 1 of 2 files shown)


src/service_daemon.rs
```diff
@@ -3179,11 +3180,19 @@ impl Zeroconf {
...
+                            // socket family. RFC 6762 doesn't require A queries
+                            // to come over IPv4 transport — Android's getaddrinfo
+                            // routinely sends both A and AAAA queries over its
+                            // preferred IPv6 mDNS socket and expects A records
+                            // to be answered with v4 addresses.
+                            let mut intf_addrs: Vec<IpAddr> = Vec::new();
+                            if qtype == RRType::A || qtype == RRType::ANY {
+                                intf_addrs.extend(service.get_addrs_on_my_intf_v4(intf));
+                            }
+                            if qtype == RRType::AAAA || qtype == RRType::ANY {
```
keepsimple1, line 3193: Nice fix! Just a note that now an `ANY` query received on IPv4 socket can return AAAA records as well.  I think that's OK.

keepsimple1 (commented):
Thanks for your PR. Nice fixes and a real useful enhancement! 

Yes please add a test and then move it out of Draft state. I'd keep all changes together instead of separated PRs as they are needed for solving the same use case.

keepsimple1:
@luqs1 friendly check: would you want to move forward with this PR?

luqs1:
Hi! @keepsimple1 
Yes I will, I'll make sure to add the tests shortly.

luqs1:
[Agent] Added `test_legacy_unicast_response` — it sends an A query from an ephemeral source port and asserts the reply arrives on that (un-joined) socket, which only happens if it was sent unicast. It also checks the question is echoed and the cache-flush bit is cleared. Kept everything in the one PR as you suggested.

On the `ANY`-over-IPv4 returning AAAA point — agreed that's fine. They're records the host genuinely has, and a querier asking `ANY` is asking for exactly that.

CI is green on all three platforms. Marking ready for review.

PR #472, opened 2026-05-21
Fix ScopedIp serialization with MessagePack

rootrascal:
The `serde(untagged)` attribute on Enums breaks serialization with binary formats like MessagePack.

I originally added that tag to optimize the redundant structures when serializing, but this can also be achieved when using `serde(flatten)` on a per-implementation basis.

Through my own testing using `rmp-serde`, I can confirm that it works after this change. If requested, I can write an additional test to cover this case.

EDIT: Actually not a problem with the format, seems like it's [an issue with rmp-serde](https://github.com/3Hren/msgpack-rust/issues/354).


keepsimple1 (commented):
I'm OK with the change. The only thing is that this is a breaking change. 

Let's think a bit more for a day. (My current thought is that this change is good for long term)

keepsimple1 (approved):
Let's go with it!

PR #474, opened 2026-06-19
Add `IfPredicate` for more flexible interface filtering

MAlba124:
We want to disable interfaces based on a regex, this isn't possible with the current `IfKind` API. Let me know if this feature is wanted and if the additional API is ok.

### DIFF
--- a/src/service_daemon.rs
+++ b/src/service_daemon.rs
@@ -909,7 +909,7 @@ pub enum IfKind {
 
 impl IfKind {
     /// Checks if `intf` matches with this interface kind.
-    fn matches(&self, intf: &Interface) -> bool {
+    pub(crate) fn matches(&self, intf: &Interface) -> bool {
         match self {
             Self::All => true,
             Self::IPv4 => intf.ip().is_ipv4(),
--- a/src/service_daemon.rs
+++ b/src/service_daemon.rs
@@ -905,6 +905,9 @@ pub enum IfKind {
 
     /// By interface index, IPv6 only.
     IndexV6(u32),
+
+    /// By a custom rule
+    Predicate(IfPredicate),
 }
 
 impl IfKind {
@@ -920,6 +923,7 @@ impl IfKind {
             Self::LoopbackV6 => intf.is_loopback() && intf.ip().is_ipv6(),
             Self::IndexV4(idx) => intf.index == Some(*idx) && intf.ip().is_ipv4(),
             Self::IndexV6(idx) => intf.index == Some(*idx) && intf.ip().is_ipv6(),
+            Self::Predicate(p) => p.matches(intf),
         }
     }
 }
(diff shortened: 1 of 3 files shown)


src/service_daemon.rs
```diff
@@ -905,11 +905,14 @@ pub enum IfKind {
 
     /// By interface index, IPv6 only.
     IndexV6(u32),
+
+    /// By a custom rule
+    Custom(CustomIfKind),
```
keepsimple1: Maybe a name bikesheding, but `Custom` feels too generic, how about

`Predicate(IfPredicate)`

src/service_daemon.rs
```diff
@@ -971,6 +975,25 @@ impl<T: Into<IfKind>> IntoIfKindVec for Vec<T> {
     }
 }
 
+#[derive(Clone)]
+pub struct CustomIfKind(std::sync::Arc<dyn Fn(&Interface) -> bool + Send + Sync>);
```
keepsimple1: As suggested above, `CustomIfKind` feels too generic, how about naming it: 

`IfPredicate`

src/service_info.rs
```diff
@@ -506,19 +506,8 @@ impl ServiceInfo {
     }
 
     fn is_address_supported(&self, intf: &Interface) -> bool {
+        let interface_supported = self.supported_intfs.iter().any(|i| i.matches(intf));
```
keepsimple1, line 509: Nice!

keepsimple1 (commented):
Thanks for your PR!  Overall it makes sense to me. A couple of comments inline, and can you please add a test?

src/service_daemon.rs
MAlba124: Agreed, renamed to `IfPredicate`.

src/service_daemon.rs
```diff
@@ -971,6 +975,25 @@ impl<T: Into<IfKind>> IntoIfKindVec for Vec<T> {
...
 
+#[derive(Clone)]
+pub struct IfPredicate(std::sync::Arc<dyn Fn(&Interface) -> bool + Send + Sync>);
+
+impl IfPredicate {
+    pub fn new(predicate: impl Fn(&Interface) -> bool + Send + Sync + 'static) -> Self {
+        Self(std::sync::Arc::new(predicate))
+    }
+
+    pub fn matches(&self, intf: &Interface) -> bool {
```
keepsimple1: nitpick:  we probably should make this method `pub(crate)`  instead, as it's not necessary to expose it to the public API surface yet.

src/service_daemon.rs
```diff
@@ -971,6 +975,25 @@ impl<T: Into<IfKind>> IntoIfKindVec for Vec<T> {
     }
 }
 
+#[derive(Clone)]
+pub struct IfPredicate(std::sync::Arc<dyn Fn(&Interface) -> bool + Send + Sync>);
```
keepsimple1, line 980: nitpick: we need a doc comment for this new public struct.

keepsimple1 (approved):
Looks good! A couple of new nitpicks inline. Thanks!

PR #489, opened 2026-08-08
feat: re-export Receiver error types (#488)

### LINKED ISSUE #488: Re-export `Receiver` error types
`Receiver` is already re-exported, but its associated error types are not. As a result, users who call `Receiver::recv()` must add `flume` as a direct dependency to their `Cargo.toml` just to access those error types.

In one of my projects, for example, I'd like to implement `From<RecvError>` for my own error type. Because `RecvError` isn't re-exported, I'm forced to depend on `flume` directly.

Since `Receiver` is already re-exported [here](https://github.com/keepsimple1/mdns-sd/blob/main/src/lib.rs#L192), could we also re-export its associated error types (e.g. `RecvError`, `RecvTimeoutError`, `TryRecvError`, etc.)? That would make the public API more self-contained and avoid requiring downstream crates to depend on `flume` solely for the receiver's error types.

keepsimple1:
Fixes #488 

## Problem:
`Receiver` was re-exported but its error types were not, so a caller that wanted to name `RecvError` -- to write `impl From<RecvError>` for its own error type, for example -- had to add `flume` as a direct dependency.

## Fix: 
Re-export the three receiver-side error types.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -191,7 +191,7 @@ pub use service_info::{
 /// A handler to receive messages from [ServiceDaemon]. Re-export from `flume` crate.
 pub use flume::Receiver;
 
-/// Errors returned by the receiving methods of [Receiver]. Re-export from `flume` crate.
+/// Errors returned by the receiving methods of `Receiver`. Re-export from `flume` crate.
 pub use flume::{RecvError, RecvTimeoutError, TryRecvError};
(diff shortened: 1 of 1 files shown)


Luni-4 (approved):
That's fine by me @keepsimple1! Thank you!

PR #495, opened 2026-08-20
fix: send goodbye packets under conflict-resolved names

dhavli:
## Problem

When probing renames a record after a name conflict (RFC 6762 section 9), announcements and query responses correctly use the new name via `DnsRegistry::name_changes` / `resolve_name`. But `unregister_service` still builds the TTL=0 goodbye records from the original `ServiceInfo` names.

Peers on the link only ever cached the *renamed* records, so the goodbye never matches anything in their caches. An auto-renamed instance therefore lingers in browse lists until its record TTLs expire after `unregister()` or `shutdown()`, instead of disappearing within seconds.

## Fix

Resolve the service fullname and hostname through the interface's `DnsRegistry` (the same `resolve_name` used by `prepare_announce`) when building the goodbye packet in `unregister_service`, so the goodbye carries the names peers actually cached. Interfaces without a registry entry keep the original names, as before.

## Test

Added `test_goodbye_uses_conflict_resolved_name`: it forces a conflict rename between two daemons, resolves the renamed instance from a third (browser) daemon, unregisters the renamed service, and asserts the browser receives `ServiceRemoved` for the renamed fullname well before any TTL could expire.

The test fails on current `main` (no goodbye ever observed for the renamed instance) and passes with this change. Full `cargo test` passes.

### DIFF
--- a/tests/mdns_test.rs
+++ b/tests/mdns_test.rs
@@ -1,7 +1,7 @@
 use if_addrs::{IfAddr, Interface};
 use mdns_sd::{
     DaemonEvent, DaemonStatus, HostnameResolutionEvent, IfKind, InterfaceId, IntoTxtProperties,
-    ScopedIp, ServiceDaemon, ServiceEvent, ServiceInfo, TxtProperty, UnregisterStatus,
+    RRType, ScopedIp, ServiceDaemon, ServiceEvent, ServiceInfo, TxtProperty, UnregisterStatus,
 };
 use std::collections::{HashMap, HashSet};
 use std::net::{IpAddr, Ipv4Addr, Ipv6Addr};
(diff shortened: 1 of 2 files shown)


tests/mdns_test.rs
```diff
@@ -2576,3 +2576,113 @@ fn timed_println(msg: String) {
...
+
+    // Register the second service with the same names to force a conflict.
+    let server2 = ServiceDaemon::new().expect("failed to start server2");
+    let server2_monitor = server2.monitor().unwrap();
+
+    let IpAddr::V4(ipv4) = ip_addr1 else {
+        panic!();
+    };
+    let bytes = ipv4.octets();
+    let ip_addr2 = IpAddr::V4(Ipv4Addr::new(bytes[0], bytes[1], bytes[2], bytes[3] + 1));
```
keepsimple1:
```suggestion
    let ip_addr2 = IpAddr::V4(Ipv4Addr::new(bytes[0], bytes[1], bytes[2], bytes[3] % 254 + 1));
```
nitpick: for `bytes[3]+1`,  to avoid overflow and avoid `255`.

keepsimple1 (approved):
Thank you for the fix! LGTM, with a minor comment inline.

PR #507, opened 2026-09-15
fix: preserve announcements after excluded address changes

Twister915:
Hello! I am submitting three fixes to bugs I encountered while developing a tvOS app for myself.

As far as I understand it, this fixes the third bug I encountered- when my Mac Studio (hosting a service) changed it's IPv6 address for privacy reasons (a macOS feature), my service appeared no longer available from my application.

After reviewing this patch myself, it looks as if the fix is pretty simple: skip updates about unsupported addresses (in my case, IPv6 addresses).

Happy to work with you through any review notes! Everything below this line is pure AI (gpt-6 astra)-

---

An auto-address service can stop answering discovery queries indefinitely after an address excluded by its policy appears on an existing interface. For example, an IPv4-only service is announced, then IPv6 privacy-address rotation adds a new address on the same interface.

`insert_ipaddr` rejects that address, but `add_interface` still attempts an announcement through the new address's socket family. No eligible records exist, so the attempt returns false and the shared per-interface status becomes `Probing`. No probes were created to restore `Announced`; query handling then skips the service.

Skip service updates for unsupported addresses, using the same predicate as automatic address insertion. Other services and the daemon's interface tracking continue to process the new address normally.

The regression registers a real service, completes its probes, verifies it is announced, then injects an excluded address into daemon state. It covers IPv4-only and IPv6-only policies using loopback and a private port; it does not change host network configuration. Reverting the fix makes the final assertion fail with `Probing` instead of `Announced`.

## Validation

- macOS, nightly Rust: targeted regression passes; fails with the implementation fix removed.
- All-feature library suite: 56 passed; `test_hostname_resolution_address_removed` failed. The same test fails on untouched upstream `75d1941` on this host.


src/service_daemon.rs
```diff
@@ -2175,6 +2175,12 @@ impl Zeroconf {
 
         for (_, service_info) in self.my_services.iter_mut() {
             if service_info.is_addr_auto() {
+                // An excluded address creates neither an announcement nor a
+                // probe. Do not demote an already announced service on this
+                // interface to Probing when such an address appears.
+                if !service_info.is_address_supported(intf) {
+                    continue;
+                }
```
keepsimple1, line 2183: nitpick:   the current form has very good readability. But this check duplicates with the same check within `insert_ipaddr` below. 

All things considered, I would prefer to change `insert_ipaddr` to return a boolean, something like this: 

```rust
pub(crate) fn insert_ipaddr(&mut self, intf: &Interface) -> bool {
    if self.is_address_supported(intf) {
        self.addresses.insert(intf.addr.ip());
        true
    } else {
        trace!("skipping unsupported address {} for service {}", intf.addr.ip(), self.fullname);
        false
    }
}
```

and then this call site becomes: 

```rust
if !service_info.insert_ipaddr(intf) {
    continue;
}
```

If you go with this change, please add a brief doc comment for   `insert_ipaddr` about its return value.

keepsimple1 (commented):
Thank you for the fix and the write-up. It's a good catch! 

One minor comment inline.

keepsimple1:
I've merged this PR so that we can prepare for other changes.

PR #509, opened 2026-09-15
fix: encode and send negative answers for absent hostname records

Twister915:
Hello! While developing a tvOS app which talks to a Rust service, I ran into a few bugs in this library that led to discovery issues. I had my agent help me write some patches to fix these issues.

This PR deals with implementing a section of the RFC which, supposedly, was not already implemented.

Like the other two, this was causing issues with discovering the service within the app- it would randomly disappear.

Here is the PR description my agent (gpt-6 astra) wrote for this-

---

Queries for an absent address family currently receive no answer even when the responder owns the hostname. Modern resolvers also ask for HTTPS/SVCB records that this responder does not publish. Explicit negative answers let those resolvers finish the lookup instead of waiting for unanswered queries.

Implement the restricted NSEC response described by [RFC 6762 section 6.1](https://www.rfc-editor.org/rfc/rfc6762#section-6.1):

- Encode NSEC's Next Domain Name as a DNS name, followed by bitmap window 0, the bitmap length, and its data. The previous serializer wrote raw hostname and bitmap bytes without the required wire structure.
- Recognize SVCB/HTTPS questions and answer absent A, AAAA, SVCB, and HTTPS records for announced hostnames with addresses on the queried interface.
- Build the bitmap from all announced registrations sharing that hostname on that interface. An IPv4-only registration must not deny an IPv6 address published by another registration.
- Preserve positive A/AAAA/ANY responses, existing suppression/rate limiting, and legacy-unicast TTL handling. Do not generate negatives for unowned or still-probing hostnames. The synthesized bitmap excludes the NSEC bit.

Encoding and generating these responses are grouped because emitting NSEC with the existing serializer would produce invalid packets. HTTPS/SVCB recognition is included here so this branch works independently of generic unknown-question handling.

## Validation

- macOS, nightly Rust: 21 parser tests and two loopback hostname-response regressions pass. Encoding and response regressions fail when their implementation fixes are removed.
- Coverage includes exact NSEC wire bytes and round trips, both address families, mixed HTTPS/AAAA/A questions, shared-hostname registrations, interface scope, and unowned/probing hostnames.
- All-feature library suite: 59 passed; `test_hostname_resolution_address_removed` failed, as it does on untouched upstream `75d1941` on this host.


src/service_daemon.rs
```diff
@@ -3498,6 +3505,15 @@ impl Zeroconf {
                         let service_hostname = dns_registry.resolve_name(service.get_hostname());
 
                         if service_hostname.to_lowercase() == question.entry_name().to_lowercase() {
+                            let ipv4 = service.get_addrs_on_my_intf_v4(intf);
+                            let ipv6 = service.get_addrs_on_my_intf_v6(intf);
+                            if ipv4.is_empty() && ipv6.is_empty() {
+                                continue;
+                            }
+                            hostname = Some(service_hostname);
+                            host_ttl = host_ttl.min(service.get_host_ttl());
```
keepsimple1, line 3514: Both `hostname` are `host_ttl` would be "last-writer-win".  But I think it's OK in practice.

keepsimple1 (approved):
Thanks for your PR!

